PR #100, opened 2013-03-30
delete response.content_type if assigning 204 (No Content) to response.status

amotl:
When assigning the 204 status code to a response object, the `content_type` attribute is still present due to `default_content_type = 'text/html'`.

We encountered problems with this behavior in some response handling code which does not explicitly create a `HTTPNoContent` response object having `empty_body = True` but reuses an existing regular response object.

For example, http unit tests using `WebTest` croak with:
> AssertionError: Content-Type header found in a 204 response, which must not return content.
> -- webtest/lint.py, line 447, in check_content_type

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -75,6 +75,16 @@ def test_set_response_status_str_generic_reason():
     assert res.status_code == 299
     assert res.status == '299 Success'
 
+def test_set_response_status_204():
+    # setting the response status to 204 should remove the content_type
+    req = BaseRequest.blank('/')
+    res = req.get_response(simple_app)
+    res.status = 204
+    assert res.status_code == 204
+    assert res.status == '204 No Content'
+    assert res.content_type is None
+    assert res.content_length is None
+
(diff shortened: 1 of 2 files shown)


bertjwregeer:
I merged this, however it was only after the fact that when I went to create a couple more tests, I realised that this change doesn't actually fix the issue.

Upon `__call__` being executed the `_app_iter` is still returned, since `_app_iter` doesn't get removed upon setting the status code, the data would still get returned to the caller.

As an aside, the fix above would cause the following to still break and send content when no content should be sent:

1. Set the status
2. Something sets `.body` or `.text`

Now the `content_length` will be set again. I think the fix should be in `__call__` instead.

bertjwregeer:
After looking at this a little more, there is no good way to support setting just the `status_code` and not also clearing all the appropriate headers, and making sure that no body is sent.

Instead of just setting a status code, please return a `webob.exc.HTTPNoContent` or your web frameworks equivalent (Pyramid has their own version).

PR #110, opened 2013-06-28
Added a number of missing headers to the request object.

RishiRamraj:
Added the following headers to the request object (mostly for logging purposes).
- content_language
- content_range
- content_md5
- expect
- expires
- from_email
- via
- warning

Converted the pragma attribute into a list, as per RFC 2616. This conversion may break dependent code.


bertjwregeer:
Closing due to no activity. I've created a new ticket to track which ones to add.

PR #118, opened 2013-08-27
try to not casually stringize each value in environ in WSGIHTTPException

moriyoshi:
...as such a value may be an object whose __str__ ends up with an unicode error.

### DIFF
--- a/tests/test_exc.py
+++ b/tests/test_exc.py
@@ -151,6 +151,56 @@ def test_WSGIHTTPException_generate_response():
         b'</html>' ]
     )
 
+def test_WSGIHTTPException_generate_response_w_unencodable_object_html():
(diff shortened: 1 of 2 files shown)


tseaver:
I'm looking at this PR along with #139.  I think I prefer the simplicity of #139's approach, but it does mean that the crash can still happen (only if a custom template is used).

amol-:
@tseaver I suppose that in case of a custom template that explicitly refers the invalid value, a crash would actually be wanted. Otherwise the developer will never know that the value is invalid and might shadow problems.

bertjwregeer:
I'm closing this in favor of #139 which I am pulling in for WebOb 1.6. Thanks for your contribution.

PR #138, opened 2014-03-07
Enhance MIMEAccept to support media type parameters

dobesv:
In many cases the media range parameter in the Accept header is used for versioning or other purposes.  The previous implementation of MIMEAccept didn't take media range parameters into account when matching accept headers.  This patch addresses this by parsing the media range parameters and comparing them when matching against the Accept header.

### DIFF
--- a/tests/test_acceptparse.py
+++ b/tests/test_acceptparse.py
@@ -243,13 +243,13 @@ def test_noaccept_contains():
 
 def test_mime_init():
     mimeaccept = MIMEAccept('image/jpg')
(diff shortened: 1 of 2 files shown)


bertjwregeer:
There is work undergoing this summer to fix the Accept handling within WebOb for languages, but others are going to go under the knife too.

The changes in this PR are backwards incompatible and intrusive, while I agree with the concept I am sorry but I don't think this is the way to solve this issue.

Related:
https://github.com/Pylons/webob/issues/315
https://github.com/Pylons/webob/issues/314
https://github.com/Pylons/webob/issues/256

PR #147, opened 2014-04-24
BaseRequest.from_file should not discard the body of PATCH requests.

thefunny42:


bertjwregeer:
This was fixed by this PR: https://github.com/Pylons/webob/pull/184.

PR #162, opened 2014-10-03
GetDict.extend() does not update QUERY_STRING

dairiki:
Calling `request.GET.extend([('foo', 'bar')])`  was not keeping ``environ['QUERY_STRING']`` in sync.  This meant that ``request.url``, ``request.path_qs``, &c. got out of sync too. This fixes that.

----

It also fixes some tests whose brokenness became  apparent after ``GetDict.extend`` was fixed. Issues with existing tests were:

- Some tests were putting non-string values into the GetDict.  GetDict doesn't like this. (You can't encode an integer to utf-8.)

- One test passed `GetDict.extend` an object whose `.items()` method returned a list of strings.  `Extend` (rightfully, I think) assumes that `.items()` returns a sequence of (key, value) pairs.

### DIFF
--- a/docs/news.txt
+++ b/docs/news.txt
@@ -1,6 +1,16 @@
 News
 ====
 
+Unreleased
+----------
+
+Bug Fixes
+~~~~~~~~~
+
+- Fixed a bug in ``webob.multidict.GetDict`` which resulted in the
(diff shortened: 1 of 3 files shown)


dairiki:
Thanks!

bertjwregeer:
Thank you!

Issue #171, opened 2014-11-09
Cookie value when using response.set_cookie() can be Unicode against spec


bertjwregeer:
In `test_response.py` we have the following tests:

https://github.com/Pylons/webob/blob/master/tests/test_response.py#L112-115
https://github.com/Pylons/webob/blob/master/tests/test_response.py#L686-690

Both of these set a cookie that contains a unicode value, however as my comment on issue #104: https://github.com/Pylons/webob/pull/104#issuecomment-28044314 shows this is not valid for a cookie value.

I am working on refactoring `response.set_cookie()` because of issue #166 and changed it to use `cookies.make_cookie()` which does not allow setting unicode values, since it correctly follows the spec and tries to encode the values to `ascii`.

I'm going to be breaking that feature in my next pull request, technically bringing WebOb inline with the RFC's, however I am unsure of how many things this will break, so the question is if we want to have a depreciation time.

Some suggestions: we can uri encode the value, which would bring the cookie value into spec, and this is recommended by some stackoverflow posts on the subject.

PR #172, opened 2014-11-09
Fix documentation/cookie handling in response.set_cookie()

### LINKED ISSUE #166: Set cookie incorrectly handles timedelta `expires` argument
`Response.set_cookie` is documented to have an `expires` argument that's a `datetime.timedelta` type.  However, if `max_age` is `None`, a `datetime.timedelta` type will result in an error: `TypeError: unsupported operand type(s) for -: 'datetime.timedelta' and 'datetime.datetime'`.  The guilty line is 778 in `response.py`:

```
        elif max_age is None and expires is not None:
            max_age = expires - datetime.utcnow()
```

This is WebOb 1.4.

A workaround is to use a `datetime.datetime`, which, although undocumented, works fine.

bertjwregeer:
This updates `response.set_cookie()` with proper documentation, and refactors it so that it uses `webob.cookies.make_cookie()` instead of it's own implementation, code-reuse for the win!

However, this does break backwards compatibility because we no longer allow unicode cookie values, these were technically contra-spec in the first place, and may not have been accepted by all browsers.

We also change from `key` to `name` because cookies have names. 

This fixes #166 and #171.

### DIFF
--- a/docs/news.txt
+++ b/docs/news.txt
@@ -4,6 +4,18 @@ News
 Unreleased
 ----------
 
+Backwards Incompatibilities
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- ``Morsel`` will no longer accept a cookie value that does not meet RFC6265's
+  cookie-octet specification. Upon calling ``Morsel.serialize`` a warning will
+  be issued, in the future this will raise a ``ValueError``, please update your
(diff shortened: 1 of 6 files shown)


bertjwregeer:
Superseded by https://github.com/Pylons/webob/pull/188

Issue #174, opened 2014-11-19
Host header validation, learning from Django


faassen:
I've been reading this document:

http://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

The most interesting bits in this document are the measures the Django project has taken to prevent various attacks.

In Django's `request.py`, there's a `get_host` method. It does the following:
- only respect X-Forwarded-Host if configuration says it can be respected. This relates to pull request #160 which propose to implement this.
- call `split_domain_port` of the host. This in turn checks against a `host_validation_re`. If it's a weird Host header, it'll not return a valid host or port. WebOb doesn't have this validation. Such a validation would be good as the port can be used to smuggle in things we don't want.
- there also seems to be a bit of support for IPv6 I can't detect in the WebOb source.
- after splitting, there's a validation done of the host specifically using `validate_host`. This uses a setting ALLOW_HOSTS to see whether the host is in a whitelist

In addition, I see WebOb there is a `host_port`, but `host_url` does not use it to split `HTTP_HOST`, instead it does it by itself.

Other URL generation code like `application_url` uses `host_url`. This could potentially be used to do a HTML injection attack into URLs generated by WebOb through the HOST header, if at least the URL is put into the template without HTML escaping. But perhaps that's the responsibility of the template language, not WebOb.

Should WebOb implement some of the measures taken by Django? If not, should we at least adjust the WebOb documentation to warn about possible attacks?

href:
Digging this old issue out because I think there would be some good in having Webob help secure all projects that depend on it. Since this should not have anything to do with 'HOST' or 'X-FORWARDED-*' maybe Webob could just assert that the host is valid inside the request class?

This might be a rather implicit thing to do, but it doesn't seem too unreasonable to me to have the attributes of the request class be valid and trusted. This could possibly be implemented on the 'Request' class, leaving the 'BaseRequest' as a fallback for anyone who doesn't want that.

renyidong:
> Developers often resort to the exceedingly untrustworthy HTTP Host header (_SERVER["HTTP_HOST"] in PHP).

That's not correct in WSGI. In WSGI world, the 'HOST' variable is provided by http server. e.g. `$host` from Nginx http://nginx.org/en/docs/http/ngx_http_core_module.html#var_host
So if a "good" http server is used, there shouldn't be such a exploit. 

If Waitress fail to comply the standards, fix Waitress.

bertjwregeer:
[...]

PR #175, opened 2014-11-20
A middleware that adjusts request.host and request.scheme according to the Forwarded header (rfc7239)

faassen:
As discussed on issue #174, implement the Forwarded header as a tween. 

If you know you have a trusted proxy that sets the Forwarded header, you can use this Tween to affect request.host and request.scheme in WebOb (and thus all code that generates links based on this information).

This is a more secure alternative than doing it inside of WebOb as in pull request #160, as web applications should _not_ accept the Forwarded header from untrusted sources as this opens them up to various attacks. In addition this tween also deals with the scheme bit, not just the host.

This does not support older forwarded headers like X_FORWARDED_HOST and X_FORWARDED_PROTO. We should do this in a similar tween along the same lines.


tests/test_forwarded.py
```diff
@@ -0,0 +1,215 @@
...
+def test_parse_forwarded_illegal_whitespace():
+    assert_raises(forwarded.ForwardedError,
+                  forwarded.parse, 'for =_something')
+    # a ForwardedError is a ValueError
+    assert_raises(ValueError,
+                  forwarded.parse, 'for =_something')
+    assert_raises(forwarded.ForwardedError,
+                  forwarded.parse, 'for= _something')
+    assert_raises(forwarded.ForwardedError,
+                  forwarded.parse, 'for=_something; by=192.51.100.17')
```
lpinca: Are you sure that whitespaces are not allowed after a semicolon?
According to the [RFC](http://tools.ietf.org/html/rfc7239) this header field is valid:
`Forwarded: for=192.0.2.43, for="[2001:db8:cafe::17]", for=unknown`
but are the following valid?
```
Forwarded: for=192.0.2.43 , for="[2001:db8:cafe::17]" , for=unknown
Forwarded: for=192.0.2.43 ,for="[2001:db8:cafe::17]" ,for=unknown
Forwarded: for=192.0.2.60; proto=http; by=203.0.113.43
Forwarded: for=192.0.2.60 ; proto=http ; by=203.0.113.43
Forwarded: for=192.0.2.60 ;proto=http ;by=203.0.113.43
```

faassen: Good question. Rereading the specs makes me unsure -- it has been a while
since I wrote this code. Need to study it a bit more.

lpinca: After reading the spec I'm not sure either, it's not very clear. I don't even know where to ask this.

piotr-dobrogost:
@faassen
Close? What's the reason?

faassen:
This PR has been open for 6 years, so I felt it was time to close it. Feel
free to pursue it further, though!

PR #180, opened 2014-11-25
Allow Range and Content-Range headers to contain any valid range unit

### LINKED ISSUE #177: Range header should parse Range-Units other than 'bytes'
The Range header parsing regex requires that the range be specified as `bytes`, e.g. `bytes = 10-20`
The only defined range unit is "bytes" according to [RFC2616#3.12](http://tools.ietf.org/html/rfc2616#section-3.12), but other range units are [allowed](https://tools.ietf.org/html/rfc7233#section-2.2).

My use case is to not have to maintain custom header parsing code when using the Range header for pagination requests. Currently a Range header like `items = 10-20` will not match.

It is not uncommon to see services defining pagination using a series of Range requests (see [Heroku's API](https://devcenter.heroku.com/articles/platform-api-reference#ranges), [Dojo toolkit](http://dojotoolkit.org/reference-guide/1.10/dojo/store/JsonRest.html#id7),  [Angular PaginateAnything](https://github.com/begriffs/angular-paginate-anything))

ltvolks:
- Range units other than 'bytes' may be provided as part of a Range
  or Content-Range header
- Updated **repr**() tests for Range and ContentRange to actually test
  the return value of the repr call (previous test was only asserting
  that **repr**() returned a 'truthy' value)
- Updated Range repr to return a string representation enclosed in angle
  brackets, matching the style of most other WebOb classes
  (vs. a string that may be able to be passed to `eval`)
- Fixes #177

### DIFF
--- a/tests/test_byterange.py
+++ b/tests/test_byterange.py
@@ -17,9 +17,14 @@ def test_range_parse():
     assert isinstance(Range.parse('BYTES=0-99'), Range)
     assert isinstance(Range.parse('bytes = 0-99'), Range)
     assert isinstance(Range.parse('bytes=0 - 102'), Range)
+    assert isinstance(Range.parse('items=0-99'), Range)
+    assert isinstance(Range.parse('ITEMS=0-99'), Range)
+
+def test_range_parse_invalid_ranges():
     assert Range.parse('bytes=10-5') is None
     assert Range.parse('bytes 5-10') is None
     assert Range.parse('words=10-5') is None
+    assert Range.parse('items extra=10-5') is None
 
 def test_range_content_range_length_none():
     range = Range(0, 100)
@@ -27,6 +32,12 @@ def test_range_content_range_length_none():
     assert isinstance(range.content_range(1), ContentRange)
     eq_(tuple(range.content_range(1)), (0,1,1))
     eq_(tuple(range.content_range(200)), (0,100,200))
+
+def test_range_content_range_units():
+    # Content-Range from Range should be expressed in same units
+    range = Range(0, 100, 'items')
+    content_range = range.content_range(200)
+    eq_(range.units, content_range.units)
 
 def test_range_for_length_end_is_none():
     # End is None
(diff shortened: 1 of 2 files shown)


ltvolks:
@bertjwregeer - Would you be able to review this PR at your convenience? See #177 for rationale. Thanks!

bertjwregeer:
I have been really conflicted on whether I wanted to merge this or not to be completely honest. It is not part of any of the standard specs, but at the same time I completely understand the utility and think it would help when I am building my own applications.

In reviewing this though, how does this interact with https://github.com/Pylons/webob/blob/master/webob/response.py#L1079? How does this work with sending a conditional response (WebOb conditional response that is). 

I need more time to understand the implications, and possibly even add a safeguard to the response code that the range unit is bytes.

ltvolks:
@bertjwregeer - The only range unit [defined by the spec](http://tools.ietf.org/html/rfc2616#section-3.12) is "bytes", but the grammar allows for other range units:

> range-unit       = bytes-unit | other-range-unit
> bytes-unit       = "bytes"
> other-range-unit = token

The Range Requests memo elaborates:

> Although the range request mechanism is designed to allow for
> extensible range types, this specification only defines requests for
> byte ranges.

[2.2.  Other Range Units](https://tools.ietf.org/html/rfc7233#section-2.2).

>    Range units are intended to be extensible.  New range units ought to
>    be registered with IANA, as defined in Section 5.1.
> 
>      other-range-unit = token

ltvolks:
@bertjwregeer - I do see your concern with https://github.com/Pylons/webob/blob/master/webob/response.py#L1079

The conditional app iterates over a range of actual bytes derived from `content_length`. The response will not match client expectations if an alternate range *unit* were specified. 

In other words, if the client requests `Range items=0-9`, the conditional app would return the first 10 *bytes* of the response body. Also the `Content-Range` header is set using the `content_length`, so the response implicitly assumes bytes and results in a mismatch:

`Content-Range: items 0-9/{{content_length}}`

I don't use the conditional app, so we haven't had an issue with the patch because we handle generating the `Content-Range` and response codes in our pagination helper.

One approach might be to explicitly disallow anything but "bytes" as a range unit when using the conditional app.

bertjwregeer:
I am closing this pull request for now. Until WebOb's `Content-Range` is fixed so that `conditional_response` continues to work I am not sure this is the right change to make. Please re-open if the concerns noted above are fixed.

PR #184, opened 2014-12-26
request: read body from file if the method is PATCH

jd:
Currently Request.from_file ignore the body of a request if the method
is PATCH. This fixes that by using the dictionary we have to know if a
request might have a body or not.
Same goes for as_bytes().


jd:
Ping ?

tseaver:
@bertjwregeer LGTM.

bertjwregeer:
Thanks @tseaver. I'll pull this during the sprints at the latest.

bertjwregeer:
I've backported this to the 1.4 branch. Will get a new version out soon.

PR #185, opened 2015-01-14
Allow wildcards to be used when comparing a MIMEAccept against an offer

### LINKED ISSUE #155: Accept matching against wildcard offers is not allowed
Wildcards cannot be used in 'offered' content types to search for a match:

```python
>>> from webob.acceptparse import MIMEAccept
>>> accept = MIMEAccept('text/*')

>>> 'text/plain' in accept
True

# Expect this to be True
>>> 'text/*' in accept
ValueError: The application should offer specific types, got 'text/*'

>>> accept.best_match(['text/html', 'text/plain', 'text/json'])
'text/html'

# Expect 'text/html' or 'text/plain' in the absence of quality parameter
>>> accept.best_match(['text/*', 'text/html', 'text/plain'])
ValueError: The application should offer specific types, got 'text/*'

# Expect 'text/*' as the best match
>>> accept.best_match(['text/*', 'foo/bar'])
ValueError: The application should offer specific types, got 'text/*'
```

The behavior and error WebOb provides is explicit and it uses the same [check](https://github.com/Pylons/webob/blob/1.4/webob/acceptparse.py#L320) anywhere it is performing a match:

```python
def _check_offer(offer):
    if '*' in offer:
        raise ValueError("The application should offer specific types, got %r" % offer)
```

This is not insensible, but it does cause an issue with Pyramid's [documented](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html) accept predicate handling: https://github.com/Pylons/pyramid/issues/1407

If you want to define a 'permissive' accept predicate like `text/*`, the Pyramid `AcceptPredicate` essentially boils down to:

```python
return 'text/*' in request.accept
```

`request.accept` is a WebOb `MIMEAccept` instance, so this check will always fail.

ltvolks:
Primary motivation is to address https://github.com/Pylons/pyramid/issues/1407, which would enable usage of wildcards as an accept predicate:

```python
@view_config(accept='text/*')
def view(request):
    ...
```

`Webob.acceptparse.MIMEAccept` now supports the following, where previously an exception was raised:

```python
>>> accept = MIMEAccept('text/html')
>>> 'text/*' in accept
True
>>> '*/html' in accept
True
>>> '*' in accept
True
```

- Fixes #155
- Addresses https://github.com/Pylons/pyramid/issues/1407

### DIFF
--- a/tests/test_acceptparse.py
+++ b/tests/test_acceptparse.py
@@ -271,7 +271,66 @@ def test_match():
     assert mimeaccept._match('image/*', 'image/jpg')
     assert mimeaccept._match('*/*', 'image/jpg')
     assert not mimeaccept._match('text/html', 'image/jpg')
-    assert_raises(ValueError, mimeaccept._match, 'image/jpg', '*/*')
+
+    mismatches = [
+        ('B/b', 'A/a'),
+        ('B/b', 'B/a'),
+        ('B/b', 'A/b'),
+        ('A/a', 'B/b'),
+        ('B/a', 'B/b'),
+        ('A/b', 'B/b')
+    ]
+    for mask, offer in mismatches:
+        assert not mimeaccept._match(mask, offer)
+
+
+def test_wildcard_matching():
+    """
+    Wildcard matching forces the match to take place against the type
+    or subtype of the mask and offer (depending on where the wildcard
+    matches)
+    """
+    mimeaccept = MIMEAccept('type/subtype')
+    matches = [
+        ('*/*', '*/*'),
+        ('*/*', 'A/*'),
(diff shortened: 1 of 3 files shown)


bertjwregeer:
Would you be so kind as to add some extra tests that would add coverage for the changes, that would be fantastic.

ltvolks:
Here you go... Curiously, `webob.util` only shows 91% coverage against `master` when I run coverage locally, but Travis seems happy enough.

bertjwregeer:
I'll work on any other coverage issues once I get this pulled in, thanks for updating this PR with tests. :-)

PR #187, opened 2015-03-18
Fix #186 — Use the same encoding to decode and encode.

### LINKED ISSUE #186: Bug in parse_qsl_text
If you look at the code here: https://github.com/Pylons/webob/blob/master/webob/compat.py#L103-L113

You will see something like:

    qs = qs.encode('latin-1')
    for value in qs.split('&'):
        yield value.decode('utf-8')

This is unlikely to work.

```
>>> '©'.encode('latin-1').decode('utf-8') 
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa9 in position 0: invalid start byte
```

Any ideas?

Natim:


Natim:
Well actually this has been done in order to have the same behavior as with the parse_qs in PY2, don't ask me why.

PR #189, opened 2015-03-28
Support stale-while-revalidate and stale-if-error

dstufft:
These supports the RFC 5861 extensions to cache control which allow a site to tell a HTTP cache that it's acceptable to serve a stale response while re-validating the currently cached response (``stale-if-revalidate``) and that it's acceptable to serve a stale response if an error occurs (``stale-if-error``). These both take a number of seconds for which it is acceptable to serve a stale response for.

I didn't write any tests because there doesn't seem to be any that exercise all of the various header options, if this needs tests let me know.

### DIFF
--- a/webob/cachecontrol.py
+++ b/webob/cachecontrol.py
@@ -202,6 +202,9 @@ class CacheControl(object):
     max_age = value_property('max-age', none=-1)
     s_maxage = value_property('s-maxage', type='response')
     s_max_age = s_maxage
+    stale_while_revalidate = value_property(
+        'stale-while-revalidate', type='response')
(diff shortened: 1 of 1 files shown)


bertjwregeer:
Thanks :-).

Tests will need to be written eventually for all of the header options, just to make sure that as I start making changes I don't break anything.

dstufft:
Do you want me to add them to this PR for these headers or would you prefer to just do them all at the same time?

bertjwregeer:
If you have the time to add tests and think you know where it is appropriate to add them, it is less work I have to do later. I was mainly adding it as a mental note on this PR for myself.

bertjwregeer:
Thanks!

PR #190, opened 2015-03-29
Include a patch from CPython's bug tracker to fix cgi.FieldStorage on 3.x

dstufft:
This backports a patch to Python's ``cgi.FieldStorage`` from the bug tracker to fix an issue on Python 3.x where a mutlipart message has a leading set of blank lines. I'm not sure if this is something webob wants to do or not, but I ran into this problem when working on [Warehouse](https://github.com/pypa/warehouse).

### DIFF
--- a/webob/compat.py
+++ b/webob/compat.py
@@ -129,3 +129,66 @@ if PY3: # pragma no cover
     from html import escape
 else:
     from cgi import escape
+
+
+# We only need this on Python3 but the issue was fixed in Pytohn 3.4.4 and 3.5.
+if PY3 and sys.version_info[:3] < (3, 4, 4):  # pragma no cover
+    # Work around http://bugs.python.org/issue23801
+    import cgi
+    from cgi import FieldStorage as _cgi_FieldStorage
+
+    class cgi_FieldStorage(_cgi_FieldStorage):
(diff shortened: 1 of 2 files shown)


bertjwregeer:
Is that patch in CPython going to get back-ported to the older Python 3 releases that also contain the bug?

In other words, is it worth patching it in WebOb if this is going to be patched in Python 3 directly?

dstufft:
It's backported to Python 3.4.4 and Python 3.5+, not anything else older than that. I doubt it's going to get backported to all of the various Python 3 via Linux package managers though (Like 3.4 on Ubuntu 14.04 etc).

bertjwregeer:
I'm just worried that sometime in the future if someone fixes the code in CPython we won't benefit from it due to keeping this around, and duplicating code.

I'll need some time think about how to best manage this moving forward, and whether it makes sense to add more guards other than just `PY3`.

dstufft:
I can add better guards now, when I had originally made this PR the patch hadn't yet been committed on the CPython side so I didn't know what versions it would be fixed in. If you want better guards I can add them.

bertjwregeer:
Yes please for the better guards.

dstufft:
Ok, better guards added!

bertjwregeer:
Thank you :-)

PR #191, opened 2015-04-07
Use duck-typing to set a quasi-integer response status.

perey:
This change checks for integer values when setting `response.status` by calling `int()` on the supplied value, instead of explicitly checking for a subclass of `int`.

This is helpful to me, because I'm writing code that wraps HTTP statuses in a class that defines `__int__()`. I'd like to be able to use WebOb together with this class without having to check whether the Response object accepted the status, or whether I need to call `int()` (or `str()` for that matter) and try again.

This change additionally removes the need for the "`if ' ' not in value`" check, which assumed `value` was interpretable by `int()` anyway and then performed the same string substitution as the setter for `response.status_code`.

### DIFF
--- a/webob/response.py
+++ b/webob/response.py
@@ -244,8 +244,12 @@ class Response(object):
         return self._status
 
     def _status__set(self, value):
-        if isinstance(value, int):
-            self.status_code = value
+        try:
+            code = int(value)
+        except ValueError:
+            pass
+        else:
+            self.status_code = code
             return
         if PY3: # pragma: no cover
             if isinstance(value, bytes):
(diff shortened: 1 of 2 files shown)


mmerickel:
Can you please fix the failing tests.

webob/response.py
```diff
@@ -256,11 +260,6 @@ def _status__set(self, value):
             raise TypeError(
                 "You must set status to a string or integer (not %s)"
                 % type(value))
-        if ' ' not in value:
-             try:
-                value += ' ' + status_reasons[int(value)]
-             except KeyError:
-                value += ' ' + status_generic_reasons[int(value) // 100]
```
mmerickel: We are losing some validation here I think.

`response.status = 'InvalidStatusCodeString'` will pass here and be set when it should fail, right?

bertjwregeer:
I pulled your changes, and rebased them on top of master.

bertjwregeer:
Thanks for your contribution!

PR #193, opened 2015-04-13
Use 'name' parameter for messages instead of 'attr' inside 'deprecated_property'

### LINKED ISSUE #68: deprecated_property message shows `None` as the attribute name
```
DeprecationWarning: The attribute None is deprecated: disabled starting WebOb 1.2, use cookies instead
```

It should use `name` instead of `attr` when calling  warn_deprecation

quantum-omega:
Fixes #68.


bertjwregeer:
Thank you! Committed to master.

PR #197, opened 2015-04-23
Don't add a charset if none is specified and the content-type is JSON.

### LINKED ISSUE #196: JSON content shouldn't need a UTF-8 on the content-type
Fix this issue: https://github.com/Pylons/pyramid/issues/1611#issuecomment-93073442

quantum-omega:
Fixes #196. If a charset is already specified in the content-type, I opted to leave it there. Else, no charset will be added to the content-type line when the the effective content-type is JSON.

I also reworked the content-type-checking condition make it more legible.

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -104,6 +104,7 @@ def test_set_response_status_code_generic_reason():
     assert res.status_code == 299
     assert res.status == '299 Success'
 
+
 def test_content_type():
     r = Response()
     # default ctype and charset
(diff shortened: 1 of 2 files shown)


mattdrayer:
@bertjwregeer -- this is a bit of a shot in the dark, I think, but I am wondering if you might be willing to add some context to an [Open edX component PR](https://github.com/edx/XBlock/pull/338) related to this particular webob change.

bertjwregeer:
@mattdrayer Yep, with pleasure. Will update whatsnew-1.6 to further clarify why we removed it.

mattdrayer:
Awesome, thanks much!

PR #203, opened 2015-06-10
Preserve wsgify default args

nickstenning:
wsgify objects can be created with args/kwargs at construction time:

``` python
def hello(req, name):
    return "Hello, %s!" % name
app = wsgify(hello, args=("Fred",))
```

Previously, these arguments would only be passed to the underlying
function (here, `hello`) when using the wsgify object as a callable that
accepts `(environ, start_response)`. If you used the object as a
callable taking a request (as documented) then the default args would be
discarded.

That is, continuing the above example:

``` python
req = Request.blank('/')
resp = req.get_response(app)  # => "Hello, Fred"
```

BUT

``` python
app(req)                      # => raises TypeError
```

This is perhaps most confusing when it interacts with
`wsgify.middleware`. For example:

``` python
@wsgify.middleware
def add_magic_header(req, app, value='seated ducks'):
    resp = req.get_response(app)
    resp.headers['Magic-Header'] = value
    return resp

app = add_magic_header(app, value='flying elephants')

resp1 = req.get_response(app)
resp2 = app(req)

resp1.headers['Magic-Header']  # => 'flying elephants'
resp2.headers['Magic-Header']  # => 'seated ducks'
```

(An executable example of the above [can be found here](https://gist.github.com/nickstenning/ee09674bc6dced0be64d).)

This pull request ensures that if variadic arguments or keyword arguments are
not provided in a call like

``` python
app(req)
```

then the defaults are used.

### DIFF
--- a/tests/test_dec.py
+++ b/tests/test_dec.py
@@ -82,6 +82,29 @@ class DecoratorTests(unittest.TestCase):
         test_app = wsgify(None, args=(1,))
         self.assertRaises(TypeError, self._testit, test_app, '/a url')
 
+    def test_wsgify_call_args(self):
+        resp_str = "args: %s, kwargs: %s"
+        def show_vars(req, *args, **kwargs):
+            return bytes_(resp_str % (sorted(args), sorted(kwargs.items())))
+
+        app = wsgify(show_vars, args=('foo', 'bar'), kwargs={'a': 1, 'b': 2})
+        resp = app(Request.blank('/'))
+
+        self.assertEqual(resp,
+                         bytes_(resp_str % ("['bar', 'foo']",
+                                            "[('a', 1), ('b', 2)]")))
+
+    def test_wsgify_call_args_override(self):
+        resp_str = "args: %s, kwargs: %s"
+        def show_vars(req, *args, **kwargs):
+            return bytes_(resp_str % (sorted(args), sorted(kwargs.items())))
+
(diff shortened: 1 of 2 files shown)


nickstenning:
As discussed with @mmerickel in `#pyramid`.

nickstenning:
I wrote up this patch last night, and I've been thinking about it some more.

I'm not sure this is the right thing to do.

Specifically, while the difference between `req.get_response(app)` and `app(req)` *is* confusing, the change in this patch:

- is backwards-incompatible
- makes it impossible to override the default arguments with *no arguments*

In addition, the whole story of default arguments seems a bit confusing:

- `wsgify` takes explicit `args` and `kw` params
- `wsgify.middleware` takes `**kw` only

So I'm submitting this for discussion, because it may be that the correct fix is actually some clearer documentation about when default arguments are used and when they are not. (i.e. "If you use the resulting callable as `app(req)`, then you are deemed to be overriding the default arguments").

I'd be interested to hear what others think.

bertjwregeer:
Since your last comment, have you had any other thoughts on how you expect this behave? If not I was planning on closing this PR for now. I don't have a good answer to any of your questions.

mmerickel:
I think it's surprising that the defaults are not used and so I think I'm in support of this PR.

nickstenning:
Heh. Thanks folks. :+1:

bertjwregeer:
We'll see what breaks when I do the next release. I think the fix is for the better! Thanks for contributing!

PR #208, opened 2015-07-22
Correct line numbers in DIY docs

cjford:
The line numbers in the explanations are off by one when compared to the code samples.


bertjwregeer:
Pulled into `docs/cleanup`. Thank you!

PR #210, opened 2015-07-31
Ensure that webob.exc.status_map is deterministic.

ryanpetrello:
Because multiple base classes share and inherit the same status code, the
values of `webob.exc.status_map` are *not* deterministic at import time.  For
example, for any given Python process, `webob.exc.status_map[500]` might
be equal to any of the following:

    * WSGIHTTPException
    * HTTPServerError
    * HTTPError
    * HTTPInternalServerError
    * HTTPRedirection

### DIFF
--- a/tests/test_exc.py
+++ b/tests/test_exc.py
@@ -1,52 +1,45 @@
 from webob.request import Request
 from webob.dec import wsgify
-from webob.exc import no_escape
-from webob.exc import strip_tags
-from webob.exc import HTTPException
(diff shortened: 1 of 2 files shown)


bertjwregeer:
Would you be so kind as to split out the change from importing things piecemeal from `webob.exc` to using `webob_exc` as a separate commit so that this change is easier to review please :-)

Thanks!

bertjwregeer:
Thank you!

ryanpetrello:
Great, thanks for merging!

PR #223, opened 2015-11-03
Allow content-length to be unset with app_iter

greghaynes:
When using an app_iter it is sometimes desireable to not set
content-length (if you, for example, do not know what your content
length is).

### DIFF
--- a/tests/test_byterange.py
+++ b/tests/test_byterange.py
@@ -23,7 +23,6 @@ def test_range_parse():
 
 def test_range_content_range_length_none():
     range = Range(0, 100)
-    eq_(range.content_range(None), None)
     assert isinstance(range.content_range(1), ContentRange)
     eq_(tuple(range.content_range(1)), (0,1,1))
     eq_(tuple(range.content_range(200)), (0,100,200))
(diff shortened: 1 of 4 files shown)


bertjwregeer:
This looks good. I'll write a test that covers the now missing line in `byterange.py`.

greghaynes:
It just occurred to me that this breaks with an open ended range request (where range.end is None).  I'll make a test for that too.

bertjwregeer:
I hadn't thought about that. I was going to dive into the RFC's and read up on the range headers to better understand all of the different scenarios.

greghaynes:
Ah, well there is a big rabbit hole. Another case is that you can define multiple ranges. AFAICT that is not supported at all by our current range.parse implementation.

bertjwregeer:
That whole logic is a huge mess, we are also re-creating the Request from the environ a second time just to be able to get the range that was requested because the new Response does not get the original Request.

I feel like the conditional response doesn't really belong in WebOb, but instead the web server should take the response and change it as necessary to satisfy the client. At this point we are stuck with it.

bertjwregeer:
@greghaynes Have you made any further progress on this? I'd like outside input on what people would like moving forward...

bertjwregeer:
I'm closing this for now. If you'd like to re-open with an updated implementation I'd be happy to review it.

PR #228, opened 2015-12-29
Use _MiddlewareFactory as decorator with arguments

renyidong:
This change allows _MiddlewareFactory to be used as decorator with argument:
```
@wsgify.middleware
def restrict_ip(req, app, ips):
    if req.remote_addr not in ips:
        raise webob.exc.HTTPForbidden('Bad IP: %s' % req.remote_addr)
    return app

@restrict_ip(ips=['127.0.0.1'])
@wsgify
def app(req):
      return 'hi'
```


mmerickel:
@renyidong Could you update this PR with a test case showing your intended workflow? I think the idea itself is fine!

renyidong:
@mmerickel Taking the example from http://docs.webob.org/en/latest/api/dec.html
```
@wsgify.middleware
def restrict_ip(req, app, ips):
    if req.remote_addr not in ips:
        raise webob.exc.HTTPForbidden('Bad IP: %s' % req.remote_addr)
    return app

@wsgify
def app(req):
    return 'hi'

wrapped = restrict_ip(app, ips=['127.0.0.1'])
```
I have to define an extra function and wrap it with the middleware. 

With this patch, I would be able to achieve same effect as the example at very beginning.

mmerickel:
@renyidong I understand the patch. I was asking if you could please add a test case to the codebase to ensure this keeps working in the future.

renyidong:
@mmerickel Apologies for my misinterpret. I appended test code to `test_middleware()`. Or you would prefer adding a separate test?

bertjwregeer:
A separate test would be preferred. Thanks!

renyidong:
@bertjwregeer Let me know if there is anything else you need.

mmerickel:
LGTM!

bertjwregeer:
Sorry for taking so long! I've merged this into master.

PR #230, opened 2016-01-12
Add ability to return a JSON body from an exception

### LINKED ISSUE #209: Allow for JSON Exception Bodies
I'm currently working on several projects that provide a JSON API using WebOb. Currently, however, whenever we use a `webob.exc` exception to return an error to the user (e.g., `webob.exc.HTTPBadRequest`) the body of that message is always in a content-type other than what they're expecting (HTML if they don't specify an Accept header, plain-text otherwise). There doesn't seem to be a pleasant, convenient, or simple way to make it use JSON beyond something like (the untested) following code:

```py
import string

import webob.exc


class WSGIHTTPException(webob.exc.WSGIHTTPException):
    body_template_obj = string.Template('{"code", ${status}, "message": "${body}", "title": "${title}"}'
    plain_template_obj = string.Template('{"error": ${body}}')


class HTTPBadRequest(webob.exc.HTTPBadRequest, WSGIHTTPException):
    pass


class HTTPUnauthored(webob.exc.HTTPBadRequest, WSGIHTTPException):
    pass


# etc.
```

This is particularly problematic because we have to redefine all of the exceptions we want to use to doubly inherit from our new sub-classed `WSGIHTTPException` and the original. It also doesn't handle the fact that we have to basically copy and paste [generate_response][] into our subclass so that we set the appropriate content-type header.

Is it too much to ask to either:

A) Add support for JSON response bodies in `WSGIHTTPException`s, or
B) Make `WSGIHTTPException` slightly more modular so we can only override parts we need?

[generate_response]: https://github.com/Pylons/webob/blob/7f98f694e7c1a569f53fb4085d084430ee8b2cc2/webob/exc.py#L302..L323

Thanks in advance,

sigmavirus24:
Closes #209

### DIFF
--- a/webob/exc.py
+++ b/webob/exc.py
@@ -165,6 +165,7 @@ References:
 
 """
 
+import json
 from string import Template
 import re
 import sys
@@ -250,7 +251,7 @@ ${body}''')
     empty_body = False
 
     def __init__(self, detail=None, headers=None, comment=None,
-                 body_template=None, **kw):
+                 body_template=None, json_formatter=None, **kw):
         Response.__init__(self,
                           status='%s %s' % (self.code, self.title),
                           **kw)
@@ -265,6 +266,8 @@ ${body}''')
         if self.empty_body:
             del self.content_type
             del self.content_length
+        if json_formatter is not None:
+            self.json_formatter = json_formatter
 
     def __str__(self):
         return self.detail or self.explanation
@@ -300,6 +303,17 @@ ${body}''')
         return self.html_template_obj.substitute(status=self.status,
                                                  body=body)
 
+    def json_formatter(self, body, status, title, environ):
+        return {'message': body,
+                'code': status,
+                'title': title}
+
+    def json_body(self, environ):
+        body = self._make_body(environ, no_escape)
+        jsonbody = self.json_formatter(body=body, status=self.status,
+                                       title=self.title, environ=environ)
+        return json.dumps(jsonbody)
+
     def generate_response(self, environ, start_response):
         if self.content_length is not None:
             del self.content_length
@@ -308,6 +322,9 @@ ${body}''')
         if accept and 'html' in accept or '*/*' in accept:
             content_type = 'text/html'
             body = self.html_body(environ)
+        elif accept and 'json' in accept:
+            content_type = 'application/json'
+            body = self.json_body(environ)
         else:
             content_type = 'text/plain'
             body = self.plain_body(environ)
--- a/webob/exc.py
+++ b/webob/exc.py
@@ -170,6 +170,7 @@ from string import Template
 import re
 import sys
 
+from webob.acceptparse import Accept
 from webob.compat import (
     class_types,
     text_,
@@ -318,11 +319,14 @@ ${body}''')
         if self.content_length is not None:
             del self.content_length
         headerlist = list(self.headerlist)
-        accept = environ.get('HTTP_ACCEPT', '')
-        if accept and 'html' in accept or '*/*' in accept:
+        accept_value = environ.get('HTTP_ACCEPT', '')
+        accept = Accept(accept_value)
+        match = accept.best_match(['application/json', 'text/html',
+                                   'text/plain'], default_match='text/plain')
+        if match == 'text/html':
             content_type = 'text/html'
             body = self.html_body(environ)
-        elif accept and 'json' in accept:
+        elif match == 'application/json':
             content_type = 'application/json'
             body = self.json_body(environ)
         else:
--- a/tests/test_exc.py
+++ b/tests/test_exc.py
@@ -1,3 +1,5 @@
+import json
+
 from webob.request import Request
 from webob.dec import wsgify
 from webob import exc as webob_exc
@@ -118,6 +120,21 @@ def test_WSGIHTTPException_html_body_w_comment():
         ' </body>\n'
         '</html>'
        )
+
+def test_WSGIHTTPException_json_body_no_comment():
+    class ValidationError(webob_exc.WSGIHTTPException):
+        code = '422'
+        title = 'Validation Failed'
+        explanation = 'Validation of an attribute failed.'
+
+    exc = ValidationError(detail='Attribute "xyz" is invalid.')
+    body = exc.json_body({})
+    eq_(json.loads(body), {
+        "code": "422 Validation Failed",
+        "title": "Validation Failed",
+        "message": "Validation of an attribute failed.<br /><br />\nAttribute"
+                   ' "xyz" is invalid.\n\n',
+    })
 
 def test_WSGIHTTPException_generate_response():
     def start_response(status, headers, exc_info=None):
--- a/tests/test_exc.py
+++ b/tests/test_exc.py
@@ -136,6 +136,42 @@ def test_WSGIHTTPException_json_body_no_comment():
                    ' "xyz" is invalid.\n\n',
     })
 
+def test_WSGIHTTPException_respects_application_json():
+    class ValidationError(webob_exc.WSGIHTTPException):
+        code = '422'
+        title = 'Validation Failed'
+        explanation = 'Validation of an attribute failed.'
+    def start_response(status, headers, exc_info=None):
+        pass
+
+    exc = ValidationError(detail='Attribute "xyz" is invalid.')
+    resp = exc.generate_response(environ={
+        'wsgi.url_scheme': 'HTTP',
+        'SERVER_NAME': 'localhost',
+        'SERVER_PORT': '80',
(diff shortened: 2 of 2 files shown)


webob/exc.py
```diff
@@ -308,6 +322,9 @@ def generate_response(self, environ, start_response):
         if accept and 'html' in accept or '*/*' in accept:
             content_type = 'text/html'
             body = self.html_body(environ)
+        elif accept and 'json' in accept:
```
bertjwregeer: If you have an XHR request that also still has `*/*` in it, this case will not get hit. Is this acceptable? Or do we want to see if JSON is listed first before checking for `HTML`/`*/*`?

bertjwregeer:
If you could please make sure that coverage is back to 100% I will accept this PR.

mmerickel:
I would really prefer a solution that parsed the accept and used `best_match` if we are going down this route of caring about accept.

```python
accept = webob.accept.Accept.parse(environ.get['HTTP_ACCEPT'])
match = accept.best_match('text/html', 'application/json')
...
```

bertjwregeer:
We currently already care about `accept`, in that it checks if HTML exists or not. Using `best_match` would indeed be a better solution.

bertjwregeer:
This kind of goes hand in hand with: https://github.com/Pylons/pyramid/pull/1378 in that we could use very similar to almost exactly the same code in both codebases since Pyramid's HTTPExceptions are a strict superset of the WebOb ones.

sigmavirus24:
@bertjwregeer I've addressed the feedback thus far. Travis is running, but just wanted to let you know this should be ready for review again (I've 100% coverage locally.)

mmerickel:
Do you have a need for passing `json_formatter` into `__init__`? That smells like feature creep to me.

sigmavirus24:
@mmerickel it's mostly to give people a way to change the "template" for their JSON response but I guess they can just subclass and override that? I was trying to provide functionality similar to `body_template` and `text_template`. I don't know if that's desired though.

sigmavirus24:
@mmerickel @bertjwregeer any further thoughts on `json_formatter` or this PR?

bertjwregeer:
I personally don't mind adding the `json_formatter`. I think it might be a good idea that allows customization and would be helpful in my own projects so I don't have to sub-class anything.

bertjwregeer:
Unless I hear a complaint I'll merge this in the next day or so.

mmerickel:
LGTM, I just wanted to force someone to defend it before it goes in. Conforming to the analogous `text_template` etc is enough for me!

sigmavirus24:
Yeah, we just can't use templates for JSON unfortunately. There are keywords to `json.dumps` that people may want to pass in the future too, but by default this should be enough. It might be fair, however, to ask people to subclass to pass those specifically. I don't need them, but I can imagine someone might, thoughts?

bertjwregeer:
@sigmavirus24 If someone wants to pass particular keywords, then they can sub-class, otherwise we can keep going and ultimately WebOb will have full JSON parsing/validation/everything else ;-)

sigmavirus24:
:ok_hand:

sjmc7:
As an additional vote for this, it'd be really helpful for the openstack projects that use webob and generally work with JSON requests and responses.

ttripp:
As an OpenStack developer, I can definitely agree that this would help keep us from having to re-implement this code.

bertjwregeer:
As an OpenStack user, I don't want you guys re-implementing this code =)

sigmavirus24:
Thanks @bertjwregeer! Let me know if I can help out with 1.6. It's clear a bunch of people would benefit from this. (I suspect @sjmc7 and @ttripp would also be willing to help out too.)

bertjwregeer:
@sigmavirus24 appreciate that, I have tagged issues for 1.6 release with the milestone 1.6, they are visible here: https://github.com/Pylons/webob/milestones/Version%201.6

All of them are currently PR's that just need someone to verify A. that they fix the issue at end, and B. that they don't cause any backwards incompatible changes (and if it does, is there a way to do it in a BW-compat way?).

Even something as simple as pulling them down, verifying they work correctly, and then adding a +1 to the PR would be very helpful!

sjmc7:
For what it's worth, i tested our service against the current master (bbca6e8ec1ae98501725817c315ca48aed5ea7fe) and everything seems to work as expected, though we don't use a lot of the features in that list. In particular, the change sigmavirus24 made seems to work as expected.

sigmavirus24:
@sjmc7 can you pull the other changes in the [1.6 milestone](https://github.com/Pylons/webob/milestones/Version%201.6) and test your service against that all merged into master?

sjmc7:
Yep, I pulled the current master, not just this patch.

sigmavirus24:
@sjmc7 what about merging the pull requests in that milestone to master in your clone (this has already been merged). You could at least verify those don't seem to break backwards compataibility.

sjmc7:
Ah, I'm sorry, i misunderstood. Been using gerrit for too long. I'll give them a try.

sjmc7:
Ok, tested with all the patches in the 1.6 milestone and don't have any regressions. Thanks for setting me straight.

bertjwregeer:
Awesome, good to know. I'm slowly going through them now and merging them, as well as doing the necessary work to document the changes.

bertjwregeer:
@sjmc7 @sigmavirus24 Released 1.6.0a0, please test on projects and let me know if there are any issues :-).

Thanks for all the help guys!

sjmc7:
Done some manual testing and run our tests on 1.6.0a0, and didn't find any problems (though as i say, we don't exercise a huge number of features). Thanks both of you for getting this done!

sigmavirus24:
I'll try to run some other project's tests against 1.6.0a0 this weekend. Thanks @bertjwregeer

sigmavirus24:
So apparently the decision in this PR that decides on returning HTML or Plain text has caused some minor backwards incompatibility. Running tests with [glance](/openstack/glance) and 1.6.0a0 showed that our (overly specific) tests were receiving plain text instead of HTML in some cases. (I haven't debugged further to determine why though.)

Also @flaper87 should be able to let us know if he saw any other test failures.

bertjwregeer:
@sigmavirus24 Before I ship 1.6.0, are there any changes that need to happen, or is this something that needs to be fixed in the Glance tests?

sigmavirus24:
Both? So I'm not sure 1.6.0 should ship with a (minor) backwards incompatibility although I still haven't looked at the failing test cases that @flaper87 found. I also think that Glance's tests (as I hinted at above) are a bit too ... specific and are testing things they shouldn't be.

sigmavirus24:
That said, if you're content with 1.6.0 having what seems like it might be a minor backwards incompatibility, you could always fix that with 1.6.1 if anyone else runs into trouble with it.

mmerickel:
I personally think it's a little silly to assume that webob can never change the format of the bundled exceptions. The library provides ways to override the format among other things if that's mission-critical. Most tools will ignore the content anyway and focus on the status code. Anyway it seems to me the change is very minor and acceptable.

sigmavirus24:
@mmerickel changing the format is one thing. Returning a totally different content-type is another, I'd think. Again, I'm only arguing this because this was clearly the result of my change. I'm not going to stop y'all from releasing 1.6.0 though.

bertjwregeer:
The content-type returned should be based upon what the original requestor asked for. If this is not the case, then things need to be changed, but if it is correct currently and the tests are failing because WebOb is now arguably doing the correct thing, then that's something we need to take into account.

FlaPer87:
I didn't see other tests break other than the ones I reported. Those tests, as @sigmavirus24 mentioned are extremely specific and they could be changed. @bertjwregeer When is the release scheduled? I'd appreciate a couple of days for us to fix them before the release.

bertjwregeer:
Is the behaviour that WebOb has correct? That's what I want to make sure of first and foremost. If you guys are expecting a different response and that is broken with the new code, why is that the case? Is the Accept matching not accurate?

bertjwregeer:
@FlaPer87 @sigmavirus24 @sjmc7 Are you guys okay with the changes moving forward?

I noticed that the tests that have changed were expecting HTML, when their accept headers stated they wanted JSON (which this change fixes for the better). Are we good to ship this?

FlaPer87:
@bertjwregeer We are, those tests have been fixed. Thanks a bunch for waiting on us and following-up! You just won a cake: :cake:

sjmc7:
Yep, we're good with it too, thanks!

sigmavirus24:
:shipit:

bertjwregeer:
Cheers! Will ship soon.

ttripp:
\O/

PR #242, opened 2016-04-14
Bugfix: exc json response now uses MIMEAccept

bertjwregeer:
This does contain a backwards incompatible change in that it changes the default back to `text/html` for WebOb exceptions compared to the default of `application/json` that set in #230. Might need to push this fix back to 1.7.0.


bertjwregeer:
@sigmavirus24 Could you please verify that none of the changes made here break glance/openstack. Since you should be explicitly requesting `application/json` I don't think it will.

I did revert to having `text/html` be the default returned for `Accept: */*`, rather than `application/json`.

sigmavirus24:
Uh, I'm not really able to verify that easily at the moment but @flaper87, @mriedem, and @ttripp should be able to help with that. :tada:

PR #249, opened 2016-04-25
Re #248: Use URL quoting to encode the Location header

### LINKED ISSUE #248: Response header "Location" should use urlencode
When you assign a unicode value to any response header, webob tries to encode it with latin-1 (https://github.com/Pylons/webob/blob/master/webob/descriptors.py#L145). But URLs often are utf-8 or worse, so that breaks when they are used in redirects in the `Location` header. If I'm reading the RFCs right (see e.g. https://stackoverflow.com/a/7654605), the `Location` header should use urlencode instead of latin-1.

I'm willing to come up with a PR for this, but I'd like someone more knowledgeable to confirm that's the way to go first. :)

wosc:
This fixes #248

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 import zlib
 import io
 
@@ -915,6 +916,15 @@ def test_location():
     assert req.get_response(res).location == 'http://localhost/test.html'
     res.location = '/test2.html'
     assert req.get_response(res).location == 'http://localhost/test2.html'
+
+def test_location_uses_urlencode():
+    res = Response()
+    res.location = u'/ümläut.html'
+    assert res.location == '/%C3%BCml%C3%A4ut.html'
+    res.location = u'http://höst:80/ümläut.html?query=föö'
+    assert (
+        res.location ==
+        'http://h%C3%B6st:80/%C3%BCml%C3%A4ut.html?query=f%C3%B6%C3%B6')
 
 def test_request_uri_http():
     # covers webob/response.py:1152
--- a/webob/descriptors.py
+++ b/webob/descriptors.py
@@ -125,8 +125,13 @@ def deprecated_property(attr, name, text, version): # pragma: no cover
         '<Deprecated attribute %s>' % name
     )
 
+def encode_latin1(value):
+    if isinstance(value, text_type) and not PY3:
+        return value.encode('latin-1')
+    else:
+        return value
 
-def header_getter(header, rfc_section):
+def header_getter(header, rfc_section, encode=encode_latin1):
     doc = header_docstring(header, rfc_section)
     key = header.lower()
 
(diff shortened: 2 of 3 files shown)


tests/test_response.py
```diff
@@ -916,6 +917,15 @@ def test_location():
...
     assert req.get_response(res).location == 'http://localhost/test2.html'
 
+def test_location_uses_urlencode():
+    res = Response()
+    res.location = u'/ümläut.html'
+    assert res.location == '/%C3%BCml%C3%A4ut.html'
+    res.location = u'http://höst:80/ümläut.html?query=föö'
+    assert (
+        res.location ==
+        'http://h%C3%B6st:80/%C3%BCml%C3%A4ut.html?query=f%C3%B6%C3%B6')
```
bertjwregeer: The host portion here should be IDNA encoded not URL encoded.

webob/descriptors.py
```diff
@@ -125,8 +127,21 @@ def fdel(self):
         '<Deprecated attribute %s>' % name
     )
 
+def encode_latin1(value):
+    if isinstance(value, text_type) and not PY3:
+        return value.encode('latin-1')
+    else:
+        return value
+
+def encode_location(value):
```
bertjwregeer: For full URL's, this should parse the URL and encode the host portion using IDNA, and the rest of the URL using url encode.

mmerickel:
I'm pretty concerned about having webob urlencode the location. It's not valid to encode an already-urlencoded value, and I'm sure plenty of people out there right now are already encoding their locations properly.

wosc:
@mmerickel Clients _not_ having to encode things by themselves is the whole point ;-) ... but yeah, that'll be backwards-incompatible.

mmerickel:
Your only chance at all to preserve compat would be to unquote the content prior to quoting it to avoid double quotes I think.

bertjwregeer:
I'm worried that if we unquote, and the user has passed a URL that looks URL encoded, but actually isn't, that we break their URL's. While in Python it isn't possible due to the WSGI spec, certain application endpoints for example allow URL encoding the `/` and treat it differently than a direct `/` in the URL (think Ruby for example).

If we unquote and re-quote we may break their URL.

For an example that won't work on Python due to WSGI, but will on other languages: https://github.com/Pylons/pyramid/issues/1008

mmerickel:
Yes that's a good point and another reason not to change the current behavior.

wosc:
I understand the backwards compatibility concerns. But I really had hoped that webob, as a "http abstraction" library, would be able insulate its clients from having to deal with encoding troubles and offer `unicode` as the interface. So this is somewhat disappointing.

At any rate, I'll abort this attempt now, since its _whole point_ is to change the current behavior, and I really don't see a backwards-compatible approach.

bertjwregeer:
It'd be okay if we didn't have backwards incompatible concerns, or concerns that we would end up breaking existing URL's that have worked in the past.

Abstraction library or not... if we change the interface we could introduce changes to someones URL structure that could mean there are interoperability concerns with other websites.

mmerickel:
One option is to only encode the url if it's unicode as that's a case that's obviously a bug right now. If it's bytes pass it untouched.

Issue #246, opened 2016-04-17
[experimental] Same-Site Cookies


bertjwregeer:
https://tools.ietf.org/html/draft-west-first-party-cookies-07

bertjwregeer:
This was merged.

PR #255, opened 2016-06-13
Support the SameSite field in Cookies

dstufft:
See https://tools.ietf.org/html/draft-west-first-party-cookies-07 and https://www.chromestatus.com/feature/4672634709082112 for more information.

This is shipping in Chrome since version 51, Opera since 39, positive support in Firefox.

### DIFF
--- a/tests/test_cookies.py
+++ b/tests/test_cookies.py
@@ -72,6 +72,18 @@ def test_cookie_httponly():
     c[b'foo'].httponly = True
     assert c.serialize() == 'foo=bar; HttpOnly'
 
+def test_cookie_samesite_strict():
+    c = cookies.Cookie()
+    c[b"foo"] = b"bar"
+    c[b"foo"].samesite = b"Strict"
+    assert c.serialize() == "foo=bar; SameSite=Strict"
+
+def test_cookie_samesite_lax():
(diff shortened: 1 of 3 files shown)


dstufft:
If this is something WebOb is interested in having, let me know and I can write tests for it.

bertjwregeer:
I added an open ticket here: https://github.com/Pylons/webob/issues/246

dstufft:
hurf durf, I am good at computer.

So I guess the question is, are you happy shipping a draft or do you want to wait until the RFC is finalized? If you're OK shipping a draft, is there any sort of remark we should make about it being a draft?

bertjwregeer:
I would also want to add a page to documentation that documents what if any draft specs are implemented.

dstufft:
This now has tests and was added to ``CookieProfile`` and invalid values raise a ``ValueError`` instead of silently working.

I still need to surface the documentation you wanted, but I'm not sure where you want the comment that this is a draft and the link to be put. Looking at the documentation, I'd *guess* you want it as part of the docstrings in the relevant methods where the ``samesite`` kwarg has been added, is that correct?

bertjwregeer:
Closing for #316

PR #257, opened 2016-06-26
Adding explicit Sphinx reference to request and response documentation

miohtama:
This way Pyramid document can correctly cross reference.


stevepiercy:
Thanks!

PR #259, opened 2016-06-29
Adds Reponse.has_body.

martinth:
This is a result of the discussion in [https://github.com/Pylons/pyramid/issues/2625](Pyramid issue 2625).

Purpose of this property is to allow users to check if the `Response` has a body without forcing an evaulation of the underlying `app_iter`. This is handy if you wan't to create a streaming response where the `app_iter` itself is lazy (i.e. streams data over a socket) or the `app_iter` can only read once.

It's pretty much the same code as @bertjwregeer suggested but I added it as property rather than es method because I figured that it's fits better in the overall class design this way. The relevant changes for Pyramid will be done later.

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -437,6 +437,37 @@ def test_app_iter_range_inner_method():
     res = Response(app_iter=FakeAppIter())
     assert res.app_iter_range(30, 40), ('you win', 30 == 40)
 
+def test_has_body():
+    empty = Response()
+    assert not empty.has_body
+
+    with_list = Response(app_iter=['1'])
+    assert with_list.has_body
+
+    with_empty_list = Response(app_iter=[b''])
+    assert not with_empty_list.has_body
+
+    with_body = Response(body='Seomthing')
+    assert with_body.has_body
+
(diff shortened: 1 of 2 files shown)


bertjwregeer:
:+1:

mmerickel:
This looks great. Just to bikeshed for a quick second... `response.has_body` could be construed to mean something like whether the status code says it should have a body or not. So I'm just curious if there's a more accurate name for this property that we could come up with. So far all I've got is... `is_body_empty`?

martinth:
You have a point. It could indeed cause confusion if you only look at the property name (which is something I totally have done in the past). `is_body_empty` is probably the most obvious choice for an alternative but I don't think it's a bad one.

bertjwregeer:
I'd be okay with the rename. I could see the confusion happening...

mmerickel:
`is_body_set` ?

bertjwregeer:
There is currently no checking if the user attempts to send a Response with an HTTP code that does not allow for a body. This is something that I may want to change in the future...

I am okay with either one. `is_body_set` or `has_body` both work for me. I'll let it simmer for a little bit longer and then go ahead and merge it.

Issue #130, opened 2014-01-17
Response wrongly keeps charset in Content-Type header


wosc:
```
>>> response = webob.Response()
>>> response.content_type = 'image/jpeg'
>>> response.headers['Content-Type']
'image/jpeg; charset=UTF-8'
```

I guess the reason is that the `content_type` property preserves parameters if the new value does not contain any, so the charset is not removed from the header when setting a content type without parameters later on.

bertjwregeer:
This was fixed in #261. Keeping any Content-Type parameters will be removed in WebOb 1.9.

PR #261, opened 2016-07-09
Fix: content_type/charset handling

bertjwregeer:
- Response.content_type removes charset unless the new content_type is a text like type that has a charset parameter
- Reverted change that was made in #253 whereby Response.charset would not allow the user to set a charset if the content type was JSON. I am not a big fan of trying to stop the user from shooting themselves in the foot, and would rather let them add/remove as needed.
- json.dumps/loads are now always UTF-8. Python returns a string to us, and we can encode it as we see fit.
- `Response.__init__` has been cleaned up to remove a lot of extraneous branch conditions that were only complicating the logic.

### DIFF
--- a/webob/response.py
+++ b/webob/response.py
@@ -112,7 +112,7 @@ class Response(object):
         else:
             self._headerlist = headerlist
         content_type = content_type or self.headers.get('Content-Type') or \
-                self.default_content_type
+            self.default_content_type
         if 'Content-Type' not in self.headers:
             self.headers['Content-Type'] = content_type
         charset = kw.get('charset')
@@ -153,7 +153,6 @@ class Response(object):
                 raise TypeError(
                     "Unexpected keyword: %s=%r" % (name, value))
             setattr(self, name, value)
-
 
     @classmethod
     def from_file(cls, fp):
@@ -220,7 +219,6 @@ class Response(object):
             headerlist=self._headerlist[:],
             app_iter=app_iter,
             conditional_response=self.conditional_response)
-
 
     #
     # __repr__, __str__
@@ -290,8 +288,7 @@ class Response(object):
             self._status = '%d %s' % (code, status_generic_reasons[code // 100])
 
     status_code = status_int = property(_status_code__get, _status_code__set,
-                           doc=_status_code__get.__doc__)
-
+                                        doc=_status_code__get.__doc__)
 
     #
     # headerslist, headers
@@ -332,7 +329,6 @@ class Response(object):
         self._headers = None
 
     headers = property(_headers__get, _headers__set, doc=_headers__get.__doc__)
-
 
     #
     # body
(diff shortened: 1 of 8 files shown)


bertjwregeer:
Closes https://github.com/Pylons/webob/issues/130 and https://github.com/Pylons/webob/issues/237

bertjwregeer:
Closes https://github.com/Pylons/webob/issues/236

bertjwregeer:
This change breaks Pyramid because it treats JSON as text, and attempts to get/set `.text` on the `Response` object.

cdent:
I reckon a common way in which this change may surprise people is if they have tests which were written to expect (the incorrect) `application/json; charset=UTF-8`. With these changes they'll now get (the correct) `application/json`. Since such tests are wrong™ I reckon that's okay.

bertjwregeer:
Any changes made in this area were going to have this issue, I'd argue that I'd rather have WebOb do the right thing rather than continue sending an invalid content-type.

bertjwregeer:
@cdent Wanted to add a little more of a note: in WebOb 1.6 there's already been some work to remove `charset=UTF-8` from certain responses that accidentally contained it, so I don't think it will be that big of a surprise to people.

CHANGES.txt
```diff
@@ -15,4 +23,38 @@ Bugfix
...
+  to the previous behaviour.
+
+  .. code::
+
+     resp = Response(charset=None, content_type='text/plain')
+     resp.default_encoding = None
+
+     # Will raise
+     try:
+        print(res.text)
```
bertjwregeer: Typo.

mmerickel:
Is there any planned deprecation warnings for this feature? Or just ripping the bandaid off?

bertjwregeer:
@mmerickel What part of it? The content-type/charset stuff is fixing the remainder of the items that should have been fixed in 1.6 but were inadequately completed.

The default_encoding is something I am not happy with, and is a change I am no longer planning to make.

webob/response.py
```diff
@@ -656,17 +766,33 @@ def _content_type__set(self, value):
...
+                warn_deprecation(
+                    'Preserving Content-Type parameters. In the '
+                    'future upon changing the Content-Type no paramaters '
+                    'will be preserved.', 1.9, 1)
+                params = self.content_type_params
+                self.headers['Content-Type'] = value
+
+                if 'charset' in params:
+                    if not _content_type_has_charset(value):
+                        warn_deprecation(
```
bertjwregeer: This should maybe be turned into a RuntimeWarning instead.

bertjwregeer:
The only backwards incompatible change is https://github.com/Pylons/webob/pull/261/files#diff-84ba6a58e29e169ddf6578bceca65dc9R768. Whereby if you currently have a content-type with a charset, and you then you replace it with one that doesn't, it is explicitly removed, but the rest of the parameters stick around.

I don't see a way to gracefully deprecate the removal of charset, since that is the crux of the issue. For the other parameters it can continue as is for now, and full removal of all of that can be done at a later point in time.

bertjwregeer:
Thanks @stevepiercy. I need to make a couple minor changes to this, and then this will be ready to go.

I forgot to document `app_iter` for the constructor so I need to do that.

PR #268, opened 2016-07-26
encget fails with invalid parameter.

Natim:
Refs https://github.com/Pylons/pyramid/pull/2725


Natim:
Do you guys think we should use ignore in that case?

tests/test_request.py
```diff
@@ -1603,9 +1602,14 @@ def test_encget_with_encattr(self):
             val = b'\xc3\xab'.decode('latin-1')
         else:
             val = b'\xc3\xab'
-        inst = self._makeOne({'a':val})
+        inst = self._makeOne({'a': val})
         assert inst.encget('a', encattr='url_encoding') == text_(b'\xc3\xab', 'utf-8')
 
+    def test_encget_does_raises_on_invalid_utf_8(self):
```
leplatrem: this test title does not seem to match the content of the test

Natim: What do you mean? It tests that if we provide a invalid utf-8 character in the PATH it ignores it right?

stevepiercy:
@Natim to save time, run `tox` locally to ensure tests will pass and docs will build.

Coverage is too low https://travis-ci.org/Pylons/webob/jobs/147479294#L349

See https://github.com/Pylons/webob/blob/master/contributing.md

Natim:
If we go for ignore we won't have a coverage problem.

tests/test_request.py
bertjwregeer: The test name states:

"encget raises when it encounters invalid UTF-8"

however there is no raising at all within the test.

bertjwregeer:
Anything that gets pulled into WebOb should have full coverage. Ignoring an entire file is not an option.

That being said, I don't think it's a good idea to just escape the characters, even with surrogates. I would much rather change UnicodeDecodeErrors into URLDecodeErrors or something along those lines.

The logic should be:

try latin-1, if that fails, try utf-8, if that fails, URLDecodeError is raised (for anything related to the URL). There was some discussion regarding that in #161.

Natim:
> try latin-1, if that fails, try utf-8, if that fails, URLDecodeError is raised (for anything related to the URL). There was some discussion regarding that in #161.

Sound like a plan :)

Natim:
I don't know enough of webob but I feel like encget is not used for ``URLDecode`` only.
Should we raise a RequestDecodeError there and use the URLDecodeError in Pyramid?

Natim:
@bertjwregeer btw are you on IRC #pyramid@freenode? What is your nickname there?

webob/request.py
```diff
@@ -98,6 +98,21 @@ def __repr__(self):
     'iso-8859-1', 'iso8859_1', 'iso_8859_1', 'iso8859', '8859',
     )
 
+
+class RequestDecodeError(UnicodeDecodeError):
+    """
+    subclass of :class:`~UnicodeDecodeError
+
+    This indicates that the server received an invalid URL.
```
Natim: An invalid Request.

webob/request.py
```diff
@@ -161,9 +176,12 @@ def encget(self, key, default=NoDefault, encattr=None):
             if not encattr:
                 return val
             encoding = getattr(self, encattr)
-            if encoding in _LATIN_ENCODINGS: # shortcut
+            if encoding in _LATIN_ENCODINGS:  # shortcut
                 return val
-            return bytes_(val, 'latin-1').decode(encoding)
+            try:
+                return bytes_(val, 'latin-1').decode(encoding)
```
bertjwregeer: The default for encoding here is `UTF-8` as defined by `Request.url_encoding` (`encget` is used only by `environ_decoder`)

https://github.com/Pylons/webob/blob/master/webob/request.py#L313

What I would like to see is if an attempt to decode using `UTF-8` fails, that we fall back to trying to decode using `latin-1`.

WebOb expects that requests are only be submitted as `UTF-8`, the HTTP standard states that the only valid encoding is `ISO-8859-1`, so I don't think it's correct to treat everything as if it is `UTF-8`.

bertjwregeer:
Just a note:

This function (`encget`) is only used for `Request.path_info` and `Request.script_name`.

When calling `Request.get` the `REQUEST_STRING` is pulled out of the environment and the passed to `parse_qsl_text` in `webob.compat`. That is where more implicit `UTF-8` decoding is done. So if you attempt to pull query parameters out using WebOb you will still blow up with a `UnicodeDecodeError`.

@Natim That's what I meant on IRC with it's more complicated than it looks. There are multiple paths where explicit encoding/decoding happens and it's not all in one place with proper fallbacks.

----

Side note, the entire change-set will have to squashed to logical commits before I accept it.

webob/request.py
Natim: I didn't change the existing behavior but I can add a test to make sure of it.

Natim:
> There are multiple paths where explicit encoding/decoding happens and it's not all in one place with proper fallbacks.

Do not worry we are not in a hurry and I will take the time needed to fix it on any path.
I am going to add tests for what you've suggest and see if I can fix it at the `webob.descriptors` level.

bertjwregeer:
Closing as I won't be accepting this. Please stay tuned, this issue is on my mind as I am working on modernizing the code base and removing Py2 support.

PR #281, opened 2016-09-26
Remove headers length from content part if present (python issue 27777)

julienmeyer:
Apply python issue https://bugs.python.org/issue27777

### DIFF
--- a/webob/compat.py
+++ b/webob/compat.py
@@ -139,7 +139,8 @@ if PY3 and sys.version_info[:3] < (3, 4, 4):  # pragma no cover
 
     class cgi_FieldStorage(_cgi_FieldStorage):
         # This is taken exactly from Python 3.5's cgi.py module, and patched
-        # with the patch from http://bugs.python.org/issue23801.
+        # with the patch from http://bugs.python.org/issue23801 and
+        # the patch from https://bugs.python.org/issue27777.
         def read_multi(self, environ, keep_blank_values, strict_parsing):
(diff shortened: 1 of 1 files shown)


bertjwregeer:
The stacktrace included with that Python issue doesn't show `read_multi` in it at all... how does removing the content-length header from the environ affect how it parses multi-part messages?

bertjwregeer:
Are you instead referring to this issue: https://bugs.python.org/issue24764?

julienmeyer:
You are right but the first issue has also the same type of problem.

In my case, I'm using Retrofit (Android application) and Pyramid on the back-end. 
Retrofit set by default the content-length on each part of the form content and when this field is set, cgi process the data as bytes and not str...

With Python 3.5 the problem is solved because content-length header field is removed by the read_multi function from cgi module.

bertjwregeer:
The first issue is separate from the issue I linked. As in, if you are hitting the first issue the patch you provided does not actually fix the issue.

julienmeyer:
After taking a look on the retrofit request sent by my application (a mix of files and string comments), my problem comes from issue 24764 and not 27777 as I have first think. 
Sorry for the mistake.

bertjwregeer:
I submitted a patch to 27777 which actually fixes both 24764 and 27777. I will add that fix to WebOb, since no Python version has this fix yet.

bertjwregeer:
I've created PR #294 for this issue. Thanks for bringing this to my attention.

PR #286, opened 2016-09-30
Performance: Response

### LINKED ISSUE #285: performance degradation on trunk
I thought I'd try Morepath with webob trunk. It's a lot slower than before. But I'll demonstrate with Pyramid,  as I don't want to confuse the issue by using development versions of Morepath. I use this benchmark:

https://github.com/faassen/howareyou

When I run Pyramid against webob 1.6.1:

```
$ howareyou -b hello -f pyramid
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       4051   24687      74     58
```

When I run Pyramid against webob trunk:

```
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       7410   13495     144     73
```

Here is the top profile functions with WebOb 1.6.1:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   100000    0.661    0.000    3.921    0.000 router.py:66(handle_request)
   100000    0.427    0.000    0.646    0.000 response.py:87(__init__)
   100000    0.340    0.000    4.744    0.000 router.py:179(invoke_subrequest)
   100000    0.239    0.000    0.456    0.000 traversal.py:632(__call__)
   100000    0.238    0.000    0.285    0.000 response.py:1043(_abs_headerlist)
   300000    0.236    0.000    0.273    0.000 interface.py:518(__hash__)
   100000    0.219    0.000    5.828    0.000 run.py:46(<lambda>)
   100000    0.204    0.000    5.524    0.000 router.py:227(__call__)
   100000    0.170    0.000    0.767    0.000 urldispatch.py:80(__call__)
   200000    0.152    0.000    0.249    0.000 decorator.py:39(__get__)
```

With master:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   100000    0.910    0.000    5.402    0.000 response.py:176(__init__)
   700000    0.731    0.000    0.883    0.000 headers.py:15(__getitem__)
   100000    0.688    0.000    8.905    0.000 router.py:66(handle_request)
   700000    0.586    0.000    1.469    0.000 _abcoll.py:379(get)
   300000    0.477    0.000    0.715    0.000 headers.py:43(__setitem__)
   300000    0.372    0.000    1.505    0.000 response.py:762(_content_type__get)
   100000    0.366    0.000    9.808    0.000 router.py:179(invoke_subrequest)
  1000000    0.302    0.000    0.511    0.000 response.py:432(_headers__get)
   100000    0.268    0.000    0.808    0.000 response.py:731(_charset__set)
   100000    0.254    0.000    0.801    0.000 response.py:796(_content_type__set)
```

Various response header manipulation code now shows up in the profile whereas before it did not.

What's interesting is that if I run the profiler against the development version of Morepath, I see almost the same webob stuff dominate the benchmarks in the same order, while usually the webob stuff is actually not in the top 10.

bertjwregeer:
Based on comments made in #285 I took a look at seeing what could be done to get some of the speed back in the synthetic test that `howareyou` provides.

In that ticket there is a patchset that brings it back to within 2k rps in the benchmark. However the maintenance burden goes up drastically, and I don't think the additional burden is worth the cost.

The new and updated `Response.__init__` is more correct, and has more code re-use by calling various properties/functions rather than trying to do it all. There is a sacrifice, but ultimately WebOb and it's `Response` object are not going to be the bottleneck, especially because outside of the synthetic test that attempts to return a `Response` object as fast as possible user code is going to dominate in cases where WebOb is used.

----

Update: After talking about the way that `Response.__init__` works with @mmerickel , a bunch of mis-features that used to exist have been removed:

If the user provides a headerlist, Response.__init__ will no longer attempt to set the default Content-Type, nor will __init__ add a charset to any existing Content-Type.

The Content-Type will only be set if no headerlist is provided, and the HTTP status code allows for a body. The charset will be set only if there is no charset on the Content-Type provided.

    resp = Response(headerlist=[], content_type='text/plain', charset='UTF-8')

will have a headerlist that is: `resp.headerlist == [('Content-Length', '0')]`

    resp = Response(status='204 No Content')

will have a headerlist that is: `resp.headerlist == []`

`__init__` will now also ignore body/app_iter if the status code has no body (100-199, 204, 304)

    resp = Response(status='204 No Content', app_iter=[b'test'])

Will not have a `Content-Length`, and `resp.body` will return `b''`. If you would like to violate HTTP standards, you may of course do:

    resp = Response(status='204 No Content')
    resp.body = b'test'

`__init__` won't let you create an invalid response, what you do with the returned response however is up to you.

Closes #285

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -1268,3 +1268,8 @@ def test_explicit_charset():
     res = Response(charset='UTF-16')
     assert res.content_type == 'text/html'
     assert res.charset == 'UTF-16'
+
+def test_set_content_type():
+    res = Response(content_type='application/json')
+    res.content_type = 'application/foo'
+    assert res.content_type == 'application/foo'
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -1288,3 +1288,27 @@ def test_set_content_type():
     res = Response(content_type='application/json')
     res.content_type = 'application/foo'
     assert res.content_type == 'application/foo'
+
+def test_raises_no_charset():
+    with pytest.raises(TypeError):
+        Response(content_type='image/jpeg', body=text_(b'test'))
+
+def test_raises_none_charset():
+    with pytest.raises(TypeError):
+        Response(
+            content_type='image/jpeg',
+            body=text_(b'test'),
+            charset=None)
+
+def test_doesnt_raise_with_charset_content_type_has_no_charset():
+    res = Response(content_type='image/jpeg', body=text_(b'test'), charset='utf-8')
+    assert res.body == b'test'
+    assert res.content_type == 'image/jpeg'
+    assert res.charset is None
+
+def test_content_type_has_charset():
+    res = Response(content_type='application/foo; charset=UTF-8', body=text_(b'test'))
+    assert res.body == b'test'
+    assert res.content_type == 'application/foo'
+    assert res.charset == 'UTF-8'
+    assert res.headers['Content-Type'] == 'application/foo; charset=UTF-8'
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -1182,20 +1182,35 @@ def test_decode_content_gzip():
     res.decode_content()
     assert res.body == b'abc'
 
-def test__abs_headerlist_location_with_scheme():
-    res = Response()
-    res.content_encoding = 'gzip'
-    res.headerlist = [('Location', 'http:')]
-    result = res._abs_headerlist({})
-    assert result, [('Location' == 'http:')]
+def test__make_location_absolute_has_scheme_only():
+    result = Response._make_location_absolute(
+        {
+            'wsgi.url_scheme': 'http',
+            'HTTP_HOST': 'example.com:80'
+        },
+        'http:'
+    )
+    assert result == 'http:'
 
-def test__abs_headerlist_location_no_scheme():
-    res = Response()
-    res.content_encoding = 'gzip'
-    res.headerlist = [('Location', '/abc')]
-    result = res._abs_headerlist({'wsgi.url_scheme': 'http',
-                                  'HTTP_HOST': 'example.com:80'})
-    assert result == [('Location', 'http://example.com/abc')]
+def test__make_location_absolute_path():
+    result = Response._make_location_absolute(
+        {
+            'wsgi.url_scheme': 'http',
+            'HTTP_HOST': 'example.com:80'
+        },
+        '/abc'
+    )
+    assert result == 'http://example.com/abc'
+
+def test__make_location_absolute_already_absolute():
+    result = Response._make_location_absolute(
+        {
+            'wsgi.url_scheme': 'http',
+            'HTTP_HOST': 'example.com:80'
+        },
+        'https://funcptr.net/'
+    )
+    assert result == 'https://funcptr.net/'
 
 def test_response_set_body_file1():
     data = b'abc'
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -18,7 +18,7 @@ def teardown_module(module):
 
 def simple_app(environ, start_response):
     start_response('200 OK', [
-        ('Content-Type', 'text/html; charset=utf8'),
+        ('Content-Type', 'text/html; charset=UTF-8'),
         ])
     return ['OK']
 
@@ -28,7 +28,7 @@ def test_response():
     assert res.status == '200 OK'
     assert res.status_code == 200
     assert res.body == "OK"
-    assert res.charset == 'utf8'
+    assert res.charset == "UTF-8"
     assert res.content_type == 'text/html'
     res.status = 404
     assert res.status == '404 Not Found'
@@ -158,7 +158,7 @@ def test_cookies():
     r2 = res.merge_cookies(simple_app)
     r2 = BaseRequest.blank('/').get_response(r2)
     assert r2.headerlist == [
-        ('Content-Type', 'text/html; charset=utf8'),
+        ('Content-Type', 'text/html; charset=UTF-8'),
         ('Set-Cookie', 'x=test; Path=/'),
         ]
 
@@ -475,19 +475,6 @@ def test_has_body():
     messing_with_privates = Response()
     messing_with_privates._app_iter = None
     assert not messing_with_privates.has_body
-
-def test_content_type_in_headerlist():
-    # Couldn't manage to clone Response in order to modify class
-    # attributes safely. Shouldn't classes be fresh imported for every
-    # test?
-    default_content_type = Response.default_content_type
-    Response.default_content_type = None
-    try:
-        res = Response(headerlist=[('Content-Type', 'text/html')], charset='utf8')
-        assert res._headerlist
-        assert res.charset == 'utf8'
-    finally:
-        Response.default_content_type = default_content_type
 
 def test_str_crlf():
     res = Response('test')
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -976,6 +976,7 @@ def test_cache_control_get():
 
 def test_location():
     res = Response()
+    res.status = '301'
     res.location = '/test.html'
     assert res.location == '/test.html'
     req = Request.blank('/')
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -1228,3 +1228,43 @@ def test_cache_expires_set_zero_then_nonzero():
     assert not res.cache_control.no_store
     assert not res.cache_control.must_revalidate
     assert res.cache_control.max_age == 1
+
+def test_default_content_type():
+    class NoDefault(Response):
+        default_content_type = None
+
+    res = NoDefault()
+    assert res.content_type is None
+
+def test_default_charset():
+    class DefaultCharset(Response):
+        default_charset = 'UTF-16'
+
+    res = DefaultCharset()
+    assert res.content_type == 'text/html'
+    assert res.charset == 'UTF-16'
+    assert res.headers['Content-Type'] == 'text/html; charset=UTF-16'
+
+def test_header_list_no_defaults():
+    res = Response(headerlist=[])
+    assert res.headerlist == [('Content-Length', '0')]
+    assert res.content_type is None
+    assert res.charset is None
+    assert res.body == b''
+
+def test_204_has_no_body():
+    res = Response(status='204 No Content')
+    assert res.body == b''
+    assert res.content_length is None
+    assert res.headerlist == []
+
+def test_204_app_iter_set():
+    res = Response(status='204', app_iter=[b'test'])
+    assert res.body == b''
+    assert res.content_length is None
+    assert res.headerlist == []
+
+def test_explicit_charset():
+    res = Response(charset='UTF-16')
+    assert res.content_type == 'text/html'
+    assert res.charset == 'UTF-16'
--- a/webob/response.py
+++ b/webob/response.py
@@ -1226,6 +1226,23 @@ class Response(object):
         if set_content_md5:
             self.content_md5 = md5_digest
 
+    @staticmethod
+    def _make_location_absolute(environ, value):
+        if SCHEME_RE.search(value):
+            return value
+
+        new_location = urlparse.urljoin(_request_uri(environ), value)
+        return new_location
+
+    def _abs_headerlist(self, environ):
+        # Build the headerlist, if we have a Location header, make it absolute
+        return [
+            (k, v) if k.lower() != 'location'
+            else (k, self._make_location_absolute(environ, v))
+            for (k, v)
+            in self._headerlist
+        ]
+
     #
     # __call__, conditional_response_app
     #
@@ -1236,26 +1253,14 @@ class Response(object):
         """
         if self.conditional_response:
             return self.conditional_response_app(environ, start_response)
+
         headerlist = self._abs_headerlist(environ)
+
         start_response(self.status, headerlist)
         if environ['REQUEST_METHOD'] == 'HEAD':
             # Special case here...
             return EmptyResponse(self._app_iter)
         return self._app_iter
-
-    def _abs_headerlist(self, environ):
-        """Returns a headerlist, with the Location header possibly
-        made absolute given the request environ.
-        """
-        headerlist = list(self.headerlist)
-        for i, (name, value) in enumerate(headerlist):
-            if name.lower() == 'location':
-                if SCHEME_RE.search(value):
-                    break
-                new_location = urlparse.urljoin(_request_uri(environ), value)
-                headerlist[i] = (name, new_location)
-                break
-        return headerlist
 
     _safe_methods = ('GET', 'HEAD')
 
@@ -1268,7 +1273,9 @@ class Response(object):
         * Range               (406 Partial Content; only on GET, HEAD)
         """
         req = BaseRequest(environ)
+
         headerlist = self._abs_headerlist(environ)
+
         method = environ.get('REQUEST_METHOD', 'GET')
         if method in self._safe_methods:
             status304 = False
--- a/webob/response.py
+++ b/webob/response.py
@@ -212,6 +212,26 @@ class Response(object):
         else:
             self._headerlist = headerlist
 
+        # Set the encoding for the Response to charset, so if a charset is
+        # passed but the Content-Type does not allow for a charset, we can
+        # still encode text_type body's.
+        # r = Response(
+        #   content_type='application/foo',
+        #   charset='UTF-8',
+        #   body=u'somebody')
+        # Should work without issues, and the header will be correctly set to
+        # Content-Type: application/foo with no charset on it.
+
+        encoding = None
+        if charset is not _marker:
+            encoding = charset
+
+        # Does the status code have a body or not?
+        code_has_body = (
+            self._status[0] != '1' and
+            self._status[:3] not in ('204', '205', '304')
+        )
+
         # We only set the content_type to the one passed to the constructor or
         # the default content type if there is none that exists AND there was
         # no headerlist passed. If a headerlist was provided then most likely
@@ -220,35 +240,57 @@ class Response(object):
         #
         # Also allow creation of a empty Response with just the status set to a
         # Response with empty body, such as Response(status='204 No Content')
-        # without the default content_type being set
+        # without the default content_type being set (since empty bodies have
+        # no Content-Type)
+        #
+        # Check if content_type is set because default_content_type could be
+        # None, in which case there is no content_type, and thus we don't need
+        # to anything
 
-        encoding = None
-        code_has_body = (
-            self._status[0] != '1' and
-            self._status[:3] not in ('204', '205', '304')
-        )
-        if headerlist is None and code_has_body:
-            content_type = content_type or self.default_content_type
+        content_type = content_type or self.default_content_type
 
-            # Check if content_type is set because default_content_type could
-            # be None, in which case there is no content_type
-            if content_type:
+        if headerlist is None and code_has_body and content_type:
+            # Set up the charset, if the content_type doesn't already have one
 
-                # Set up the charset, if the content_type doesn't already have one
+            has_charset = 'charset=' in content_type
 
-                has_charset = 'charset=' in content_type
+            # If the Content-Type already has a charset, we don't set the user
+            # provided charset on the Content-Type, so we shouldn't use it as
+            # the encoding for text_type based body's.
+            if has_charset:
+                encoding = None
 
-                if not has_charset and charset is not _marker and charset is not None:
-                    encoding = charset
-                    content_type += '; charset=' + charset
-                elif not has_charset and charset is _marker and self.default_charset:
-                    # Optimize for the default_content_type case, avoid a function call
-                    if (
-                        content_type is self.default_content_type or
-                        _content_type_has_charset(content_type)
-                    ):
-                        encoding = self.default_charset
-                        content_type += '; charset=' + self.default_charset
+            # Do not use the default_charset for the encoding because we
+            # want things like
+            # Response(content_type='image/jpeg',body=u'foo') to raise when
+            # trying to encode the body.
+
+            new_charset = encoding
+
+            if (
+                not has_charset and
+                charset is _marker and
+                self.default_charset
+            ):
+                new_charset = self.default_charset
+
+            # Optimize for the default_content_type as shipped by
+            # WebOb, becuase we know that 'text/html' has a charset,
+            # otherwise add a charset if the content_type has a charset.
+            #
+            # Even if the user supplied charset explicitly, we do not add
+            # it to the Content-Type unless it has has a charset, instead
+            # the user supplied charset is solely used for encoding the
+            # body if it is a text_type
+
+            if (
+                new_charset and
+                (
+                    content_type == 'text/html' or
+                    _content_type_has_charset(content_type)
+                )
+            ):
+                content_type += '; charset=' + new_charset
 
             self._headerlist.append(('Content-Type', content_type))
 
--- a/webob/response.py
+++ b/webob/response.py
@@ -212,9 +212,6 @@ class Response(object):
         else:
             self._headerlist = headerlist
 
-        # Set up the content_type
-        content_type = content_type or self.default_content_type
-
         # We only set the content_type to the one passed to the constructor or
         # the default content type if there is none that exists AND there was
         # no headerlist passed. If a headerlist was provided then most likely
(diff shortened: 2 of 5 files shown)


bertjwregeer:
```
# master before this branch
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       9430   10604     144     73

# Fixes applied
(howareyou)alexandra:howareyou xistence$ howareyou -b hello -f pyramid
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       8112   12327     123     72
```

bertjwregeer:
```
# WebOb 1.6.1
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       5299   18870      74     58

# WebOb with this branch
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       5051   19797      71     56
```

bertjwregeer:
@mmerickel Could you take a hard look at this and make sure I didn't miss anything. I added a bunch of new tests because while everything was "covered" there was no tests for `default_body_encoding` and `default_charset`

bertjwregeer:
@faassen this one is for you.

bertjwregeer:
```
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       4829   20706      69     54
```

Cleaned up `_abs_headerlist`. 69 tcalls...

```
   Ordered by: internal time
   List reduced from 65 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   100000    0.838    0.000    4.670    0.000 router.py:65(handle_request)
   100000    0.589    0.000    0.704    0.000 response.py:181(__init__)
   100000    0.427    0.000    5.776    0.000 router.py:166(invoke_subrequest)
   100000    0.318    0.000    7.016    0.000 run.py:42(<lambda>)
   100000    0.297    0.000    0.418    0.000 response.py:1241(__call__)
   100000    0.295    0.000    0.567    0.000 traversal.py:632(__call__)
   300000    0.264    0.000    0.313    0.000 interface.py:518(__hash__)
   100000    0.250    0.000    6.600    0.000 router.py:214(__call__)
   100000    0.214    0.000    0.945    0.000 urldispatch.py:80(__call__)
   200000    0.197    0.000    0.323    0.000 decorator.py:34(__get__)
```

WebOb is listed twice. `Response.__init__` and `Response.__call__`.

bertjwregeer:
It may be possible to cut down the `__call__` more, AFAIK the following status code are allowed to have a `Location` header on them:

- 201
- 301
- 302
- 303
- 305
- 307

Sending a `Location` header on any others is not allowed, and from bug reports across the internet it causes issues with many clients that would be on the receiving end of that.

webob/response.py
```diff
@@ -225,43 +222,66 @@ def __init__(self, body=None, status=None, headerlist=None, app_iter=None,
...
-        if (
-            self.content_type is None and
-            headerlist is None and
-            _code_has_body(self.status_code)
-        ):
-            self.content_type = content_type
+        encoding = None
+        code_has_body = (
+            self._status[0] != '1' and
+            self._status[:3] not in ('204', '304')
```
bertjwregeer: Add `205` to this small list. It is like a `204` but it requires the client to reset the document view (i.e. clear a form).

webob/response.py
```diff
@@ -225,43 +222,66 @@ def __init__(self, body=None, status=None, headerlist=None, app_iter=None,
...
 
-        # Set up app_iter
-        if app_iter is None:
+        # Set up app_iter if the HTTP Status code has a body
+        if app_iter is None and code_has_body:
             if isinstance(body, text_type):
-                encoding = self.charset
+                # Fall back to trying self.charset if encoding is not set. In
+                # most cases encoding will be set to the default value.
+                encoding = encoding or self.charset
```
mmerickel: Shouldn't `default_body_encoding` be used here before raising an error?

webob/response.py
```diff
@@ -225,43 +222,66 @@ def __init__(self, body=None, status=None, headerlist=None, app_iter=None,
...
+
+                if not has_charset and charset is not _marker and charset is not None:
+                    encoding = charset
+                    content_type += '; charset=' + charset
+                elif not has_charset and charset is _marker and self.default_charset:
+                    # Optimize for the default_content_type case, avoid a function call
+                    if (
+                        content_type is self.default_content_type or
+                        _content_type_has_charset(content_type)
+                    ):
```
mmerickel: Does the `default_content_type` imply that it needs a charset?

webob/response.py
```diff
@@ -1248,7 +1270,15 @@ def conditional_response_app(self, environ, start_response):
         * Range               (406 Partial Content; only on GET, HEAD)
         """
         req = BaseRequest(environ)
-        headerlist = self._abs_headerlist(environ)
+
+        # Build headerlist, if there is a Location header, make it absolute
+        headerlist = [
```
mmerickel: note something about how this `_abs_headerlist` code is copied around / inlined from somewhere for future editors... possibly even leaving the original impl in place with tests, and a note about where it is inlined to... and the inline points should point back to the canonical impl

tests/test_response.py
```diff
@@ -1241,3 +1243,48 @@ def test_cache_expires_set_zero_then_nonzero():
...
+    assert res.content_type is None
+
+def test_default_charset():
+    class DefaultCharset(Response):
+        default_charset = 'UTF-16'
+
+    res = DefaultCharset()
+    assert res.content_type == 'text/html'
+    assert res.charset == 'UTF-16'
+    assert res.headers['Content-Type'] == 'text/html; charset=UTF-16'
```
bertjwregeer: I am not sure I follow...

webob/response.py
bertjwregeer: It does in the case of the default_content_type we ship with, but not if it was overridden. I think I'd want to change this to `content_type is 'text/html'` since that is our default, and it saves us a function call and a bunch of checking in the default case.

webob/response.py
bertjwregeer: I don't think so, I would love to hear arguments for/against it... but `default_body_encoding` is strictly for `.text` to me. 

`__init__` won't let you create a `204 No Content` with a body for example (because setting Content-Length on such a response is not allowed by the HTTP spec), `__init__` will do it's best to give you a valid `Response`. If you make changes/set invalid options on the `Response` after the fact, then so be it.

There is no way to influence `default_body_encoding` in `__init__` so you end up using the default all of the time. I'd recommend encoding before it reaches `Response`. Changing this would also be b/w incompatible, if you set the `content_type` to `image/jpg` in older versions of WebOb you couldn't provide a text body because there would be no charset set.

Reading through the 1.6.1 code though, it seems if you set `charset` and `content_type` it would happily create `image/jpg; charset=utf-8` and then use the charset if passed in. So I guess I could use the provided `charset` as if it were the `default_body_encoding`, even-though it is not added the `content_type`.

If you do `Response(text='sometext')` then `__init__` will call the `.text` attribute and `default_body_encoding` will be used to set the `body` up though...

webob/response.py
bertjwregeer: Good idea, but:

I've been thinking about removing this entirely. Making the `Location` absolute really doesn't provide much especially since all clients since HTTP 1.0 support relative Location's. It's doing extra work that feels unnecessary.

Thoughts?

tests/test_response.py
mmerickel: I was commenting that this test confirms that if you set charset **after** content_type, it still appends the charset as expected.

webob/response.py
mmerickel: I'm not against it but it will be a library regression for a bunch of people - at least breaking a bunch of test suites until it's fixed. The behavior itself **should** work, but may expose some malformed non-browser clients or anyone handling location headers themselves.

webob/response.py
mmerickel: What about baking the charset into the `default_content_type` instead of handling them separately in this case? Then `default_content_type` is unambiguous, and `default_charset` is only used if someone provides a partial content-type.

webob/response.py
mmerickel: afaict passing `body` as a `text_type` is similar to invoking `.text` (and in my view setting `.body` with text data should be equivalent to setting `.text` with the data). It looks to me that right now `Response(body=u'foo')` works but `r = Response(); r.body = u'foo'` does not.

bertjwregeer: `body` has accepted only `bytes` (or it's equivalent `str` in Py2 for a VERY long time): https://github.com/Pylons/webob/blob/0.8/webob/__init__.py#L1445

It would be an API asymmetry to change that, because `r.body` will always return bytes, which may be unexpected. I feed it text, I get back bytes from the same property?

That's what `.text` is for.

The fact that `Response(body=u'foo')` works is that in 0.9 `__init__` was updated to call `unicode_body`: https://github.com/Pylons/webob/blob/0.9/webob/__init__.py#L1364 when the "body" provided was unicode. I assume it was a convenience thing, because `.body` was not updated to match. `.unicode_body` sometime in its history got renamed to `.text`, and instead of calling `self.unicode_body` vs `self.body` the features from it were inlined into `__init__` probably for speed reasons, which brings us back to where we are today.

Unfortunately the name "body" is being re-used for something that accepts both `bytes`/`str` and `unicode`.

----

That being said, falling back to `default_body_encoding` seems like it could be a sane thing to do, and matches what was done in the past.

Side note: Ian at one point did only set the default `charset` on text types:

https://github.com/Pylons/webob/blob/0.9/webob/__init__.py#L1357

Somewhere between that version and 1.6.x it was lost, and thus if you didn't explicitly provide a `charset` providing `u'foo'` as a `body` parameter would have failed due to not having a valid charset.

webob/response.py
bertjwregeer: It wouldn't allow someone to sub-class `Response` and override only the `default_charset` or set it to `None` for instance.

tests/test_response.py
bertjwregeer: Ah, that's not what this test checks.

This test verifies that sub-classing `Response` with a new `default_charset` actually uses the `default_charset` provided by the sub-class and not the original.

webob/response.py
bertjwregeer: WebOb 1.6.1 behaviour:

```
>>> Response(content_type='image/jpg', body=u'test',charset=None).body
  File "/Users/xistence/Projects/webob/webob/response.py", line 139, in __init__
    "You cannot set the body to a text value without a "
TypeError: You cannot set the body to a text value without a charset
>>> Response(content_type='image/jpg', body=u'test').body
  File "/Users/xistence/Projects/webob/webob/response.py", line 139, in __init__
    "You cannot set the body to a text value without a "
TypeError: You cannot set the body to a text value without a charset
>>> Response(content_type='image/jpg', body=u'test',charset='UTF-16').body
b'\xff\xfet\x00e\x00s\x00t\x00'
```
If we used `default_body_encoding` both of the first two examples would succeed and no error would be raised.

tests/test_response.py
mmerickel: Oh I guess I read that wrong. Does what I specified work?

webob/response.py
mmerickel: Sorry I don't follow. It seems to me if you trust `default_content_type` as fully specified then it simplifies things instead of asking them to work together.

webob/response.py
bertjwregeer: If we want to break those first two examples, I can make that change, so that it looks more like `.text` and falls back to `default_body_encoding`. I am -0 on it personally.

mmerickel: Okay, thanks for the clarification. I think that if `body` is unicode then the goal should be to raise. The inconsistency bothers me. There's a lot of work being put into getting things correct in the `__init__` but my experience is usually affecting the `Response` **after** init via things like `request.response.foo = bar` or just creating a response object before configuring it so if there's things that I'm losing by not doing it all at `__init__` time, I'd like to know.

tests/test_response.py
```diff
@@ -1288,3 +1288,26 @@ def test_set_content_type():
...
+            body=text_(b'test'),
+            charset=None)
+
+def test_doesnt_raise_with_charset_content_type_has_no_charset():
+    res = Response(content_type='image/jpeg', body=text_(b'test'), charset='utf-8')
+    assert res.body == b'test'
+    assert res.content_type == 'image/jpeg'
+    assert res.charset is None
+
+def test_conten_type_has_charset():
```
mmerickel: `s/conten/content`

tests/test_response.py
```diff
@@ -1288,3 +1288,26 @@ def test_set_content_type():
...
+    res = Response(content_type='application/foo; charset=UTF-8', body=text_(b'test'))
+    assert res.body == b'test'
+    assert res.content_type == 'application/foo'
+    assert res.charset == 'UTF-8'
```
mmerickel: Can I expect `assert res.headerlist = [('Content-Type', 'application/foo; charset=UTF-8')]` here? I was slightly surprised to see `res.content_type` not include the charset.

webob/response.py
bertjwregeer:
```
    class DefaultCharset(Response):
        default_charset = 'UTF-16'

    res = DefaultCharset()
    assert res.content_type == 'text/html'
    assert res.charset == 'UTF-16'
    assert res.headers['Content-Type'] == 'text/html; charset=UTF-16'
```

Would no longer change the charset of the `content_type` if `default_content_type` was set to `text/html; charset=UTF-8`. To change the `default_charset` in that case, you'd have to also override the `default_content_type`.

The test would have to become:

```
    class DefaultCharset(Response):
        default_content_type = 'text/html; charset=UTF-16'
        default_charset = 'UTF-16'

    res = DefaultCharset()
    assert res.content_type == 'text/html'
    assert res.charset == 'UTF-16'
    assert res.headers['Content-Type'] == 'text/html; charset=UTF-16'
```

tests/test_response.py
bertjwregeer:
```
r = Response()
r.content_type = 'application/foo'
r.charset = 'UTF-16'

assert r.headers['Content-Type'] == 'application/foo; charset=UTF-16'
```

However:

```
r = Response()

# Change charset to UTF-16
r.charset = 'UTF-16'

assert r.headers['Content-Type'] == 'text/html; charset=UTF-16'

# Change the content_type, this removes charset and raises a warning
r.content_type = 'application/foo'

assert r.headers['Content-Type'] == 'application/foo'
```

webob/response.py
mmerickel: In your first example, what is `default_content_type`?

tests/test_response.py
mmerickel: Wait didn't you just say that I can't set `.body` with unicode and get back bytes? So if I set `response.content_type = 'application/foo; charset=UTF-8'` then `assert response.content_type == 'application/foo'` will pass?

webob/response.py
bertjwregeer: Brought back `_abs_headerlist`, and added an optimization to the default path to only call it when the HTTP Status Code allows for a `Location` header. If someone has a `Location` header on any other status codes they are not within spec, and it won't be made absolute. If this breaks test suites, then they need to get fixed.

webob/response.py
bertjwregeer: Changing it to always raise of the `body` is unicode would break a lot of existing applications that depend on the behaviour.

mmerickel: Not interested in deprecating it? Usually you have seem happy to aim for consistency so I'm a little surprised.

bertjwregeer: There is nothing that you are losing by not doing it at `__init__` time, after `__init__` time there are however less seatbelts. For example:

```
res = Response(status='204 No Content')
res.body = b'A body'
```

Would create a response that had:

```
assert res.body == b'A body'
assert res.status == '204 No Content'
assert res.headers['Content-Length'] == 6
```

Which is very much not allowed by the HTTP spec. a `204 No Content` response is not allowed to have any message-body. This also means no `Content-Type`/`Content-Length` is allowed. `Content-Length: 0` is also invalid, and various HTTP libraries will throw the entire response away as invalid.

tests/test_response.py
bertjwregeer: I've updated these tests to make it more clear.

webob/response.py
bertjwregeer: It is the `default_content_type` from `Response`: `text/html`.

bertjwregeer:
```
    class DefaultCharset(Response):
        default_charset = 'UTF-16'

    res = DefaultCharset()
    assert res.content_type == 'text/html'
    assert res.charset == 'UTF-16'
    assert res.headers['Content-Type'] == 'text/html; charset=UTF-16'

    assert DefaultCharset.default_content_type == 'text/html'
    assert DefaultCharset.default_charset == 'UTF-16'
```

with the change you'd recommended:

```
    assert DefaultCharset.default_content_type == 'text/html; charset=UTF-8'
    assert DefaultCharset.default_charset == 'UTF-16'
```

tests/test_response.py
bertjwregeer: Yes. This change was made because of this:

```
res = Response()
res.content_type = 'image/jpg'
```

used to keep the content-type parameters, which also meant stuff like:

```
assert res.headers['Content-Type'] == 'image/jpg; charset=UTF-8'
```

This was changed in WebOb 1.6.0. Currently other content-type parameters are copied from the old to the new:

```
res = Response(content_type='application/foo; level=1; charset=UTF-8')
res.content_type = 'image/jpeg'

assert res.headers['Content-Type'] == 'image/jpeg; level=1'
```

Which is going to go away in the future. Content-Type parameters shouldn't be kept when moving from one Content-Type to another.

webob/response.py
mmerickel: Yes, if for some reason you wanted the `default_charset` to not be UTF-8 then that'd be what you do.

webob/response.py
```diff
@@ -1287,13 +1296,11 @@ def __call__(self, environ, start_response):
...
 
-        # Build headerlist, if there is a Location header, make it absolute
-        headerlist = [
-            (k, v) if k.lower() != 'location'
-            else (k, self._make_location_absolute(environ, v))
-            for (k, v)
-            in self._headerlist
-        ]
+        # If the HTTP Status code might have a Location header, try to make it absolute
+        if self._status[:3] in ('201', '301', '302', '303', '305', '307'):
```
mmerickel: For starters this doesn't support 308. In general, I don't like this change. It's actually fairly difficult to figure out what responses support Location... you're only checking right now for the responses that **require** Location. It's complicated.

webob/response.py
bertjwregeer: As for deprecating it... I'm all for it, but that means `Response('body')` no longer works. Are we ready to deal with that fall-out?

```
>>> Response('test')
  File "/Users/xistence/Projects/webob/webob/response.py", line 308, in __init__
    raise TypeError("Expected bytes value but received text.")
TypeError: Expected bytes value but received text.
```

I'm generally all for it, except this time I have a feeling it's going to break MANY applications.

mmerickel: Well in your example you didn't specify `body=` so it feels like you could set `.text` instead of `.body` since it was left ambiguous. For the case of `Response(body='body')` I can see the concern.

webob/response.py
bertjwregeer: Setting Location on a `200 OK` HTTP Status code is not allowed for example, and will cause some intermediaries to change the status code to a 302 (Apache's mod_proxy IIRC), and others will send a 500 Server Error back to the remote client (Squid). Various HTTP checkers (for example OpenStack has some neat ones) also complain loudly. You can't just set Location on stuff willy nilly. Same thing with Content-Length for example, you can't set that on responses without a body.

I can add 308 to the list, or I can drop the optimization completely. It doesn't make that big of a difference.

bertjwregeer:
```
(howareyou)alexandra:howareyou xistence$ howareyou -b hello -f pyramid
Benchmark : hello
Frameworks: pyramid
                ms     rps  tcalls  funcs
pyramid       4723   21173      70     55
```

Dropped the extra optimization, not worth it.

webob/response.py
bertjwregeer: The first parameter that `__init__` takes is body. If I made that `.text` instead, things would go badly too:

(Assumes that the first param to `__init__` is `.text`)

```
>>> Response(b'test')
  File "/Users/xistence/Projects/webob/webob/response.py", line 335, in __init__
    setattr(self, name, value)
  File "/Users/xistence/Projects/webob/webob/response.py", line 647, in _text__set
    "(not %s)" % type(value))
TypeError: You can only set Response.text to a unicode string (not <class 'bytes'>)
```

webob/response.py
bertjwregeer: Making that change doesn't have much effect on performance at all. I'd rather leave it as is.

tests/test_response.py
bertjwregeer: Yes. `r.content_type` can be set with a full `content_type` string and everything works correctly. As in `r.headers['Content-Type'] == 'application/foo; charset=UTF-8'`. `r.content_type` only returns the content-type not the content-type with parameters.

Certain properties have some really bad assymetric API's. Unfortunately they have been that way for a long time, can't change that now.

webob/response.py
mmerickel: This is a usability change. I wasn't arguing for it in favor of performance.

bertjwregeer:
@mmerickel I'm sorry, but I don't think this is the right place to discuss the API issues in `Response`.

At this point I am happy with the changes in this PR, are there any other major concerns that you have? Outside issues with the API as it stands?

webob/response.py
mmerickel:
```
def __init__(self, *args, **kw):
    if len(args) > 0:
        if instance(args[0], bytes):
            kw['body'], args = args[0], args[1:]
        elif isinstance(args[1], text_type):
            kw['text'], args = args[0], args[1:]
    if len(args) > 0:
        raise ValueError('unknown args')
```

mmerickel:
Well where do you want me to go from here? You made API changes in this PR that is labeled performance. I need to discuss them somewhere. Do you want to make a new performance PR that is **only** a performance change?

bertjwregeer:
Wait... what API change did I make? The only change made was that charset is no longer added to a Content-Type that is provided in the `headerlist`.

No other API changes have been made.

You are discussing and asking about `.body` and `.text`, as well as `.content_type` but those have not changed behaviour at all in this PR.

bertjwregeer:
```
____________________________________________________________________ test_204_has_no_body ____________________________________________________________________

    def test_204_has_no_body():
        res = Response(status='204 No Content')
        assert res.body == b''
>       assert res.content_length is None
E       assert 0 is None
E        +  where 0 = <Response at 0x102ae3390 204 No Content>.content_length

tests/test_response.py:1244: AssertionError
___________________________________________________________________ test_204_app_iter_set ____________________________________________________________________

    def test_204_app_iter_set():
        res = Response(status='204', app_iter=[b'test'])
>       assert res.body == b''
E       assert b'test' == b''
E         Left contains more items, first extra item: 116
E         Use -v to get the full diff

tests/test_response.py:1249: AssertionError
_________________________________________________ test_doesnt_raise_with_charset_content_type_has_no_charset _________________________________________________

    def test_doesnt_raise_with_charset_content_type_has_no_charset():
        res = Response(content_type='image/jpeg', body=text_(b'test'), charset='utf-8')
        assert res.body == b'test'
        assert res.content_type == 'image/jpeg'
>       assert res.charset is None
E       assert 'utf-8' is None
E        +  where 'utf-8' = <Response at 0x104b768d0 200 OK>.charset

tests/test_response.py:1278: AssertionError
=========================================================== 3 failed, 1095 passed in 5.92 seconds ============================================================
```

Test failures on `master` that are working on `perf/responses`.

mmerickel:
Well the Location header is no longer normalized for certain response codes. Honestly the rest of it is just super hard to keep track of because the original behavior isn't exactly documented with regards to `content_type`, `charset`, `default_content_type` and `default_charset`. As you showed several of the tests on master with regards to those arguments have broken, so something changed and I've been trying to figure out what the intended behavior is. Feel free to merge this if you want to discuss the API issues elsewhere, but I don't know the original behavior well enough to evaluate this with regards to just the performance optimizations (and through no fault of anyone, certainly not you, I don't have a lot of faith in simply the test suite to show what has actually changed).

bertjwregeer:
> Well the Location header is no longer normalized for certain response codes.

Nope, I removed that code. It is always normalized, for all HTTP response codes.

> Honestly the rest of it is just super hard to keep track of because the original behavior isn't exactly documented with regards to content_type, charset, default_content_type and default_charset.

The only tests that fail are the "no body" ones. Which I added (on this new branch perf/responses). `204 No Content` shouldn't have a `Content-Length` and shouldn't have a `body`.

The only charset one that fails is when the user passes in a `headerlist` with a `Content-Type` that doesn't have a `charset` already on it. We discussed this on IRC, you said if someone passes a `headerlist` then there should be no changing what is in that `headerlist` since it is a low level api.

I agreed to that.

Currently on `master` if you pass in a `headerlist` with a `Content-Type` it will add the `charset` to it automatically.

WebOb 1.6.1/webob master:

```
res = Response(headerlist=[('Content-Type', 'text/html')], charset='UTF-8')

assert res.headers['Content-Type'] == 'text/html; charset=UTF-8'
```

WebOb perf/responses:

```
res = Response(headerlist[('Content-Type', 'text/html')], charset='UTF-8')

assert res.headers['Content-Type'] == 'text/html'
```

once again, we discussed this on IRC. `headerlist` is a lower level API.

In short, only NEW tests for new behaviour I've added to make sure that behaviour can't regress in the future are failing on master. Tests from master pass without issues on perf/responses.

Those same tests that fail on `master` that are on `perf/response` also fail when testing on 1.6-branch. So there is no changes there.

If you want me to pull the "no body" changes, I can do that, and make a new branch for the couple of line changes.

> Feel free to merge this if you want to discuss the API issues elsewhere,

Yes, I don't think that this is the place.

> but I don't know the original behavior well enough to evaluate this with regards to just the performance optimizations (and through no fault of anyone, certainly not you, I don't have a lot of faith in simply the test suite to show what has actually changed).

The test suite isn't half bad at showing regressions. Most of the changes made in master for the 1.7 series so far have required changes in the test suite, and I spent a lot of time going through the tests when I converted it to py.test.

mmerickel:
Sounds good, thanks for being so patient with me. Let's discuss the content-type and charset stuff elsewhere.

bertjwregeer:
No problem. I am all for making things better. If you want to put together a bunch of questions I'll do my best to answer them and any history surrounding them.

PR #287, opened 2016-09-30
Feature: default body encoding

### LINKED ISSUE #271: Add default_body_encoding to Response
Add a new default_body_encoding, this is used for .text if there is no charset for the content-type.

This is a backwards incompatible change that breaks the assumption that setting .text on a binary content-type will raise an error.

This will allow a user to also change the default_body_encoding on the fly before setting the body using .text, thereby allowing the encoding to happen to whatever they wish.

bertjwregeer:
Pyramid uses `.text` in it's renderers, however with the changes made to remove charset when there is no charset on the content-type, this meant that things would break.

Instead add a new `default_body_encoding` to `Response` that may be used as the alternative to charset. This will allow `.text` to continue functioning.

It's a small backwards incompatible change, but if you were relying on WebOb to complain about no charset being set when accessing `.text` I'd love to have a word with you :-).

Closes #271

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -693,8 +693,13 @@ def test_body_del():
 
 def test_text_get_no_charset():
     res = Response(charset=None)
+    assert '' == res.text
+
+def test_text_get_no_default_body_encoding():
+    res = Response(charset=None)
+    res.default_body_encoding = None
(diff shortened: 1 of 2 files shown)


bertjwregeer:
@mmerickel This one's for you!

Issue #254, opened 2016-06-03
Request.set_cookie(expires) requires a UTC datetime but doesn't document this


dobesv:
Currently `Response.set_cookie` will accept a `datetime` instance to specify the time at which the cookie should expire.

However, this `datetime` instance is expected to be a `utc` datetime, for example `datetime.utcnow() + timedelta(seconds=3600)`, not a local datetime like `datetime.now() + timedelta(seconds=3600)`.

Also, the provided `datetime` instance must NOT be timezone-aware or you will get an error.  For example passing `datetime.utcnow(tz=pytz.utc) + timedelta(seconds=3600)`.

If callers are not aware of these requirements, their application might fail unexpectedly if the app moves between time zones.

bertjwregeer:
`expires` on the `Response.set_cookie` is something I am going to deprecate. Setting a `max_age` automatically calculates the expiration correctly, and the API that is used to actually make the cookie no longer supports an expiration parameter.

Things that need to get fixed for now:

- If date time is timezone aware, convert it to UTC, without TZ
- Change date time from local to UTC

WebOb 1.9:

Drop `expires` from the API entirely.

bertjwregeer:
There is no way to determine if a datetime is a local one or not. It's documented as long accepting a UTC datetime.

At this point, timezone aware datetime's are converted to UTC, and those work correctly now.

PR #292, opened 2016-11-18
Bugfix: set_cookie datetime

bertjwregeer:
This fixes the bug reported by @dobesv in #254. Also documents that `expires` should not be a localtime, and that it should not be set in the past.

### DIFF
--- a/tests/test_response.py
+++ b/tests/test_response.py
@@ -860,6 +860,36 @@ def test_set_cookie_expires_is_timedelta_and_max_age_is_None():
     assert val[2] == 'a=1'
     assert val[3].startswith('expires')
 
+def test_set_cookie_expires_is_datetime_tz_and_max_age_is_None():
+    import datetime
+    res = Response()
+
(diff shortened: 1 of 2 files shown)


mmerickel:
What is the issue with setting expires in the past? AFAIK this is a common technique for clearing a cookie.

bertjwregeer:
@mmerickel: There is `unset_cookie` for that and I would rather warn because it may mean the user accidentally passed in a local time `datetime` object and setting the cookie is not doing what they think it is going to do.

bertjwregeer:
I can change the warning message, expires is going away in the future anyway. Setting max_age to 0 is a better idea.

mmerickel:
Yeah this is a little wishy washy but it sort of feels like something webob shouldn't have an opinion on.

bertjwregeer:
If you don't think it is a good idea, I can remove that warning. I only added it late last night because of the local time vs utc time issue and maybe a warning would be nice, I have no opinion either way.

Issue #245, opened 2016-04-15
Convert from unittest.TestCase to py.test


digitalresistor:
The following files did not contain nosetests based tests, and thus were not yet converted to py.test:

```
test_client.py
test_compat.py
test_dec.py
test_multidict.py
test_static.py
test_util.py
```

PR #300, opened 2016-11-24
Bug/issue 299

julienmeyer:


tests/test_compat.py
```diff
@@ -51,7 +51,7 @@ def test_text_alternate_encoding(self):
         self.assertTrue(isinstance(result, bytes))
         self.assertEqual(result, b'La Pe\xc3\xb1a')
 
-class Test_cgi_FieldStorage_Py3_tests(object):
```
bertjwregeer: We use py.test for testing, this does not need to be sub-classed from unittest.TestCase.

bertjwregeer: The only reason the other tests in this file still use `unittest.TestCase` is due to the fact that they have not yet been converted to use `assert`.  See https://github.com/Pylons/webob/issues/245

webob/compat.py
```diff
@@ -197,6 +198,15 @@ def read_multi(self, environ, keep_blank_values, strict_parsing):
                     self.bytes_read += len(hdr_text)
                     parser.feed(hdr_text.decode(self.encoding, self.errors))
                     headers = parser.close()
+                    # Some clients add Content-Length for part headers, ignore them
+                    if 'content-length' in headers:
```
bertjwregeer: +1

I like this better than just throwing out the Content-Length data entirely. I can't help but feel like `FieldStorage` needs to be fixed so that it does the right thing even if the `Content-Length` exists vs just throwing it out. I think using the fact that something has `Content-Length` as a reason to use a binary tempfile is a bad idea.

bertjwregeer:
Thank's Julian! Would you mind adding this test case to the patch I submitted for Py27777 or would you mind if I stole your test case/patch wholesale to update the patch there?

https://bugs.python.org/issue27777

Will get this pulled in, and backported to 1.7-branch and get a new rc out! Appreciate your testing!

tests/test_compat.py
julienmeyer: No problem, I will remove it.

julienmeyer:
Hello, you can add my test case into the patch. You can also add a note to the other python bug (24764).

PR #313, opened 2017-03-27
Prevent leaking if file exists outside of static path

inytar:
This fix prevents the static.DirectoryApp from leaking that a file exists
outside of the given path by first checking if the path is correct
before testing if the file exists.


bertjwregeer:
Thanks!

PR #318, opened 2017-05-22
Replacement PR for #232

bertjwregeer:
Closes #232


bertjwregeer:
I implemented the namedtuple suggestion from @mmerickel

PR #321, opened 2017-06-10
Fix BytesWarning in webob.cookies._value_quote

JelleZijlstra:
When running Python 3 with the `-b` option, calling str() on a bytes instance raises a warning. This came up for us when running code with `-bb` enabled (which frequently helps catch bugs).


bertjwregeer:
Thank you!

PR #327, opened 2017-06-16
Document acceptparse classes

whiteroses:
This is far from complete as documentation for the acceptparse.py classes, but as a lot may change in the near future, this pull request is mainly to set out the classes that need to be documented &mdash; I'll be changing and adding to this over the summer.

I added a comment about running into bugs with `:members` and `:inherited-members`, so that we remember why we chose to use `automethod` and `autoattribute` instead; and gave the comment a date so we can come back to it at a later date perhaps to check if the bugs have been fixed. Hope that's okay.

### DIFF
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -6,12 +6,49 @@ Headers
 
 Accept-*
 ~~~~~~~~
+.. # 2017-06-16: There appears to be various bugs with the ``:members`` and
+.. # ``:inherited-members`` options, both when the two are used together, and
+.. # when one of them is combined with ``autoattribute``, so we are using
+.. # ``automethod`` and ``autoattribute`` instead.
 
 .. automodule:: webob.acceptparse
-.. autoclass:: Accept
-   :members:
+
+.. autoclass:: AcceptCharset
+
+   .. automethod:: best_match
+   .. automethod:: quality
+
+.. autoclass:: AcceptEncoding
+
+   .. automethod:: best_match
+   .. automethod:: quality
+
+.. autoclass:: AcceptLanguage
+
+   .. automethod:: best_match
+   .. automethod:: quality
+
 .. autoclass:: MIMEAccept
-   :members:
+
+   .. autoattribute:: accepts_html
+   .. automethod:: accept_html
+   .. automethod:: best_match
+   .. automethod:: quality
+
+.. autoclass:: MIMENilAccept
+
+   .. automethod:: best_match
+   .. automethod:: quality
+
+.. autoclass:: NilAccept
+
+   .. automethod:: best_match
+   .. automethod:: quality
+
+.. autoclass:: NoAccept
+
+   .. automethod:: best_match
+   .. automethod:: quality
 
 Cache-Control
 ~~~~~~~~~~~~~
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -16,16 +16,19 @@ Accept-*
 .. autoclass:: AcceptCharset
 
    .. automethod:: best_match
+   .. automethod:: parse
    .. automethod:: quality
 
 .. autoclass:: AcceptEncoding
 
    .. automethod:: best_match
+   .. automethod:: parse
    .. automethod:: quality
 
 .. autoclass:: AcceptLanguage
 
    .. automethod:: best_match
+   .. automethod:: parse
    .. automethod:: quality
 
 .. autoclass:: MIMEAccept
@@ -33,6 +36,7 @@ Accept-*
    .. autoattribute:: accepts_html
    .. automethod:: accept_html
    .. automethod:: best_match
+   .. automethod:: parse
    .. automethod:: quality
 
 .. autoclass:: MIMENilAccept
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -6,11 +6,6 @@ Headers
 
 Accept-*
 ~~~~~~~~
-.. # 2017-06-16: There appears to be various bugs with the ``:members`` and
-.. # ``:inherited-members`` options, both when the two are used together, and
-.. # when one of them is combined with ``autoattribute``, so we are using
-.. # ``automethod`` and ``autoattribute`` instead.
-
 .. automodule:: webob.acceptparse
 
 .. autoclass:: AcceptCharset
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -234,6 +234,11 @@ class AcceptCharset(Accept):
     """
     @staticmethod
     def parse(value):
+        """
+        Parse ``Accept-Charset`` header.
+
+        Return iterator of ``(charset, qvalue)`` pairs.
+        """
         latin1_found = False
         for m, q in Accept.parse(value):
             _m = m.lower()
@@ -272,6 +277,11 @@ class MIMEAccept(Accept):
     """
     @staticmethod
     def parse(value):
+        """
+        Parse ``Accept`` header.
+
+        Return iterator of ``(media range, qvalue)`` pairs.
+        """
         for mask, q in Accept.parse(value):
             try:
                 mask_major, mask_minor = [x.lower() for x in mask.split('/')]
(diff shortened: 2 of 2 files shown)


stevepiercy:
If you claim something is a bug, provide a link to it in [Sphinx's issue tracker](https://github.com/sphinx-doc/sphinx/issues/). [Perhaps it is this one](https://github.com/sphinx-doc/sphinx/issues/3739)? If there is no relevant issue reported there, then create one.

It's possible that it is not a bug, too, but how the Sphinx directives are used. A reproducible example including what you tried, what you expected, and what you observed would be very helpful toward resolving the issue.

whiteroses:
I intended to go back to it later to investigate and file the issue, if there was free time left during the GSoC period, and if not in my own time afterwards. The problem does not appear to be the issue you linked to; but I can't be certain without spending more time looking more deeply into it. I had searched Sphinx's issues tracker, and did not find anything that appeared to be the same issue. But again, to be certain takes time, and filing a decent bug report &mdash; testing all the combinations, looking through the source, trying to locate the problem and maybe fixing it myself (they have 584 open issues; I didn't feel I should just dump a bug report on them without trying to do as much of the work as I could). So I intended to make a note of it for now, and return to it at a later date, especially as I have my hands full dealing with issues for WebOb, and did not think I should spend too much of that limited time wrestling with Sphinx, when my job is to check and fix Accept handling for WebOb. (But if you'd prefer I spend the time looking into the issue for Sphinx, I can certainly do that.)

The issue was that when I tried to use `:members` and/or `:inherited-members` with explicit lists of members, in combination with each other or with `automethod`/`autoattribute`, the lists of members appeared to be ignored. I had already spent a lot of time, trying many combinations to try to pin down the problem: using the options in different orders, with or without blank lines in-between, with or without one or more of the options or directives... and the issue kept coming up. I'm certainly no Sphinx expert, and while I spent a lot of time trying to rule out the possibility, it is still possible that it was something I was doing wrong &mdash; but again, limited time on GSoC. It got to a point where I didn't think I should spend more time on it, as there was so much to do for WebOb, and I didn't think that was what you would want me to spend the time on. `automethod` and `autoattribute` appear to be doing the job just fine, and are pretty much equivalent for this use; `:members` and `:inherited-members` would only perhaps be slightly more convenient. I just thought I should make a note there for now, so someone else doesn't come along, think Oh why don't we use `:members` and `:inherited-members`, and run into the same issue. But I can easily remove the comment, and it wouldn't change anything. If you prefer, I can do that, and make a note instead in an issue to go back to it at a later date. Alternatively, if you would like me to spend this time investigating this for Sphinx, I can do that too; there is just a lot to do for WebOb, and GSoC is not that long.

stevepiercy:
If you have time, create a new branch in your repo that demonstrates the issue, say `sphinx-why-do-you-hate-me-so`, push it to GitHub repo, and share a link to that. Instant reproducible example!

If I had a reproducible example to try, I could help find a solution on your behalf, letting you focus on the functional code. We're a team.

I noticed that you write `:inherited-members` without the closing `:`. It should be `:inherited-members:` and `:members:`, in case that is what causes the issue.

whiteroses:
Thanks! I missed the closing colons on `members` and `inherited-members` when talking about them in the comments, but when writing the actual code, I followed the docs pretty closely, so it wasn't the problem, no.

I started working on a new branch as you suggested to demonstrate the issue, but quickly realised it'd be much quicker for you to type or copy&paste the changes, as the changes between steps are very small, and switching between commits/branches would likely be more confusing and take longer. So here are steps to reproduce the issues:

From the current master branch (before this pull request), add a docstring &mdash; say, "parse docstring" &mdash; to MIMEAccept.parse() in acceptparse.py. Run sphinx-build.

Why is parse() documented with docstring under Accept, but not under MIMEAccept (which overrides parse())?
May have something to do with it being a static method, but parse() appears under Accept without problems.

For MIMEAccept, we want to document parse, accept_html, accepts_html, best_match and quality.

Let's try :inherited-members: to see if it makes any difference.

    .. autoclass:: MIMEAccept
       :members:
       :inherited-members:

best_match() and quality() are now documented under MIMEAccept, but still no parse().


We also want to control which members to document, so let's try an explicit list of members with :members:.

    .. autoclass:: MIMEAccept
       :members: accepts_html
       :inherited-members:

No effect, as if there is no list.


If we remove the :inherited-members:

    .. autoclass:: MIMEAccept
       :members: accepts_html

The :members: list now works as expected, with only accepts_html showing under MIMEAccept. (Still no parse().)


Let's try removing :members: and leaving only :inherited-members:.

    .. autoclass:: MIMEAccept
       :inherited-members:

We get non-inherited *and* inherited members, but no parse().

If we feed a member name that does not exist to :members:.

    .. autoclass:: MIMEAccept
       :members: does_not_exist

we get a warning. But with :inherited-members:

    .. autoclass:: MIMEAccept
       :members: does_not_exist
       :inherited-members:

No warning this time, as if the :members: list is just ignored. (And we get members and inherited members, but no parse().)


What I originally wanted to do was something like:

    .. autoclass:: MIMEAccept
       :members: accept_html, best_match, parse

i.e. be able to control which inherited and non-inherited members to include, and autodocument overridden staticmethods like parse(). But parse() is not appearing.


Combine with automethod?

    .. autoclass:: MIMEAccept
       :members: accepts_html, best_match

       .. automethod:: parse

We get parse()! But not its docstring.


automethod works just fine with the other methods, inherited or not:

    .. autoclass:: MIMEAccept
       :members: accepts_html, best_match

       .. automethod:: parse
       .. automethod:: accept_html
       .. automethod:: quality

We get the docstrings for accept_html and quality, but not for parse.


What if we remove :members:?

    .. autoclass:: MIMEAccept

       .. automethod:: parse

Still no docstring for parse.

There is no autostaticmethod (I tried, just in case it was undocumented.)
autofunction doesn't seem to work either.


And if we try to document AcceptEncoding.parse, which is inherited but not overridden?

    .. autoclass:: AcceptEncoding
       :members: parse

We get parse! We get docstring! But oddly, we don't get the label of *static*, to indicate that it's a staticmethod.

To check that it works the same with automethod:

    .. autoclass:: AcceptEncoding

       .. automethod:: parse

Same result: parse, with docstring, but no indication that it's a staticmethod like we get under the Accept class.


So there seems to be at least two issues:
1) If we use the :inherited-members: flag, the :members: option and any list after it appears to have no effect. This is not the impression given by the docs, as far as I can see; in fact, I don't understand why the docs suggest we use :members: and :inherited-members: together, if that is the effect, and if :inherited-members: by itself appears to include non-inherited members in the output.
2) There are issues with autodocumenting a staticmethod like parse, especially if it has been overridden by the class.

This is a fraction of the things I've tried. At some point it may be easier to read the source to see what's happening. I need to get back to working on Accept-Language handling &mdash; please let me know if anything is unclear, and thanks for looking into the issues with Sphinx!

whiteroses:
I added docstrings for `.parse()` in the classes where it was overridden. The classes that inherit it without overriding, I left them to use the inherited docstring from `Accept` for now.

Added `.parse()` to documentation with `automethod`, even though as previously described, there appears to be various issues with how Sphinx treats staticmethods. Sorry I probably wasn't very clear with my last comment @stevepiercy; you can now see in this branch the issues with `.parse()` &mdash; the docstrings are not shown for `.parse()` in the classes where it is overridden (`AcceptCharset` and `MIMEAccept`), and the *static* label is missing for `.parse()` in the classes where it is not overridden (`AcceptEncoding` and `AcceptLanguage`). Hopefully my last comment now makes more sense &mdash; I was trying various combinations of `:members:` and `:inherited-members` (because `.parse()` is inherited) to fix the issues, but switched to `automethod` instead because its behaviour seems to make more sense. The issues with `.parse()` are still there though, as mentioned.

I've removed the comment, as I can see how it may not be very helpful. This pull request is only a start to fixing the documentation of Accept* handling, and a lot more needs to be done to update/flesh out many of the docstrings. This PR is mainly to have that outline of the classes in the docs, so I can start adding to them as I work on things. If everything looks okay, once it's merged I can expose `._parsed` and `._parsed_nonzero`, document them and move forward. Thanks!

mmerickel:
Thank you @whiteroses.

We can worry about inherited-members and such later, this is strictly an improvement already.

PR #332, opened 2017-06-30
Bugfix: ipv6 host support

### LINKED ISSUE #331: IPv6 support
Parts of WebOb haven't been adapted to work in IPv6 environment:
- request.domain will split IPv6 addresses incorrectly
- request.host_port will split IPv6 addresses incorrectly
- request.host_url will split IPv6 addresses incorrectly
- .. maybe more places, I haven't checked all sources

This issue is similar to #174 however there it's about mitigating potential vulnerabilities and this issue is only about making API work correctly. Software using WebOb can workaround this by parsing the Host header manually, but it's just wasteful because everybody has to repeat the same work, that should be done by WebOb (as promised by the API)

bertjwregeer:
Closes #331

### DIFF
--- a/src/webob/request.py
+++ b/src/webob/request.py
@@ -413,8 +413,8 @@ class BaseRequest(object):
         e = self.environ
         host = e.get('HTTP_HOST')
         if host is not None:
-            if ':' in host:
-                host, port = host.split(':', 1)
+            if ':' in host and host[-1] != ']':
+                host, port = host.rsplit(':', 1)
             else:
                 url_scheme = e['wsgi.url_scheme']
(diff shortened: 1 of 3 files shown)


src/webob/request.py
```diff
@@ -414,7 +414,7 @@ def host_port(self):
         host = e.get('HTTP_HOST')
         if host is not None:
             if ':' in host:
```
mmerickel: This check is not enough to determine to split it based on `:`. The IPv6 address may not have a port but still has `:` characters.

bertjwregeer: Fixed.

Issue #256, opened 2016-06-19
AcceptLanguage _match leads best_match to ignore specificity


TropicalPenguin:
Hello,

If best_match is passed:

``` python
offers=['en', 'en-GB'] # Order important
```

and there is an 'en-GB' mask with quality 1, and an 'en' mask with quality 0.8, then 'en' will be considered the best match with 'en-GB', due to:

``` python
or item == mask.split('-')[0] # From AcceptLanguage._match()
```

So I suspect there needs to be some tiebreaking logic specific to languages, to replace:

``` python
# 'text/plain' overrides 'message/*' overrides '*/*'
# (if all match w/ the same q=)
if matched_by.count('*') <= mask.count('*'):
    continue
```

bertjwregeer:
Accept handling is all kinds of special in HTTP...

But I am not sure that WebOb is wrong:

```
offers=['en', 'en-GB']
mask='en-GB, en;q=0.8'
```

It is acceptable to return `en` according to [RFC4647](https://tools.ietf.org/html/rfc4647#section-3)  `Section 3.3.1 Basic Filtering`. 

>   A language range matches a
>   particular language tag if, in a case-insensitive comparison, it
>   exactly equals the tag, or if it exactly equals a prefix of the tag
>   such that the first character following the prefix is "-".

Since both `en` and `en-GB` both have the same weight, 1 in this case, it matches the above from the RFC. Accept handling unfortunately is not entirely sane.

WebOb could implement the Extended Filtering described in 3.3.2 and I wouldn't be opposed to a patch that would implement that, so long as it doesn't break the existing implementations.

renyidong:
We may use the similar logic: https://github.com/Pylons/webob/blob/master/webob/acceptparse.py#L155

By the way, this logic doesn't seem to be correct:
Consider `a-*-c; a-b-*;` Both have same number of `*`, but when matching against `a-b-c`,  `a-b-*` should take priority, as it is more specific.

bertjwregeer:
What valid "language" has three dashes in it? IIRC `en` and `en-GB` are valid languages, there is no `en-GB-something`. This is because the first element `en` is the language code, and the second element `GB` would indicate the country code.

See https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.4

renyidong:
@bertjwregeer There are plenty examples in RFC: https://tools.ietf.org/html/rfc4647#section-3.4
[...]

PR #335, opened 2017-07-25
Accept language wip

whiteroses:
Some remaining questions:

1. How would you like me to do the imports in the tests? I can switch to module scope or fixtures, if either is your preference. (Module scope does seem to produce quite a lot of noise when one import is broken though, especially with tox and coverage, and I don't think fixtures are being used for imports in any WebOb tests at the moment.)
2. Would you prefer I switch from test classes to test functions? I see benefits to using test classes (namespacing, so e.g. we wouldn't have to repeat the class name in every test function name, leaving more room to be more descriptive with the name of the test; classes can be folded in editors), but don't know the benefits of test functions over classes. Happy to learn, and switch if test functions are your preference.
3. The `fget` in `accept_language_property`, like the existing `fget` in `accept_property`, creates a new instance every time. Is that necessary, and might it be good to cache the instance? Most people would probably not expect the header to be re-parsed every time they access `request.accept_language`.
4. I did what I think you might want with the deprecation warnings, but please let me know if you'd like me to change or remove them. Also, calling `warn()` with `DeprecationWarning` and `PendingDeprecationWarning` is now ignored by default from Python version 2.7. I saw quite a few `DeprecationWarning`s in WebOb -- is that an issue?
5. I think I can fix some of the issues in `.quality()` and `.best_match()`, so that they conform to the RFC in how they handle e.g. `q=0` and `*`, but they would still be their own unique algorithms that are not specified or mentioned in the RFCs. (And I'm pretty sure `best_match()` does not implement the matching scheme used in Apache, at least not for `Accept-Language`?) So they are probably not methods you'd want to maintain in the long run -- is it worth fixing those issues when we would like people to move away from them so you can deprecate them?
6. `AcceptLanguageValidHeader`, `AcceptLanguageInvalidHeader`, and `AcceptLanguageNoHeader` all inherit from `AcceptLanguage`, so they can all be identified as an `AcceptLanguage` header. Would it be worth making `AcceptLanguage` an abstract base class and specifying the properties and methods that can be expected in an `AcceptLanguage` (sub)class?
7. What should I do about the `modifier` parameter in `Accept.quality()` and the `default_quality` parameter in `NilAccept.quality()`? (I explained the issues with the `modifier` parameter with comments in `AcceptLanguageValidHeader.quality()` and the `default_quality` parameter is not used at all in `NilAccept.quality`, and does not appear to match any corresponding parameter in `Accept.quality()` either.)

Please let me know if there is anything else you'd like me to change. Thanks!

### DIFF
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -27,7 +27,6 @@ Accept-*
 .. autofunction:: create_accept_language_header
 
 .. autoclass:: AcceptLanguage
-   :members:
 
 .. autoclass:: AcceptLanguageValidHeader
    :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -19,6 +19,7 @@ Accept-*
 .. autofunction:: create_accept_language_header
 
 .. autoclass:: AcceptLanguage
+   :members: parse
 
 .. autoclass:: AcceptLanguageValidHeader
    :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
@@ -27,11 +28,13 @@ Accept-*
 
 .. autoclass:: AcceptLanguageNoHeader
    :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
-	     __radd__, __str__, basic_filtering, best_match, lookup, quality
+	     __radd__, __str__, parse, basic_filtering, best_match, lookup,
+             quality
 
 .. autoclass:: AcceptLanguageInvalidHeader
    :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
-	     __radd__, __str__, basic_filtering, best_match, lookup, quality
+	     __radd__, __str__, parse, basic_filtering, best_match, lookup,
+             quality
 
 .. autoclass:: MIMEAccept
    :members:
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -31,8 +31,8 @@ Accept-*
    :inherited-members:
 
 .. autoclass:: AcceptLanguageValidHeader
-   :members: __init__, __contains__, __iter__, parse, basic_filtering,
-             best_match, lookup, quality
+   :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
+             parse, basic_filtering, best_match, lookup, quality
 
    .. autoinstanceattribute:: header_value
       :annotation:
@@ -40,7 +40,8 @@ Accept-*
       :annotation:
 
 .. autoclass:: AcceptLanguageNoHeader
-   :members: __init__
+   :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
+             basic_filtering, best_match, lookup, quality
 
    .. autoinstanceattribute:: header_value
       :annotation:
@@ -48,7 +49,8 @@ Accept-*
       :annotation:
 
 .. autoclass:: AcceptLanguageInvalidHeader
-   :members: __init__
+   :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
+             basic_filtering, best_match, lookup, quality
 
    .. autoinstanceattribute:: header_value
       :annotation:
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -6,14 +6,6 @@ Headers
 
 Accept-*
 ~~~~~~~~
-
-.. # We use ``autoinstanceattribute`` instead of  ``members`` to document
-   # instance attributes for now, because of
-   # https://github.com/sphinx-doc/sphinx/issues/904
-   # (``autoattribute`` and ``autodata`` do not work with instance attributes.)
-   # As mentioned there, ``autoinstanceattribute`` is undocumented. However, it
-   # has been around for years, and works.
-
 .. automodule:: webob.acceptparse
 
 .. autoclass:: AcceptCharset
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -28,7 +28,6 @@ Accept-*
 
 .. autoclass:: AcceptLanguage
    :members:
-   :inherited-members:
 
 .. autoclass:: AcceptLanguageValidHeader
    :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -24,6 +24,8 @@ Accept-*
    :members:
    :inherited-members:
 
+.. autofunction:: create_accept_language_header
+
 .. autoclass:: AcceptLanguage
    :members:
    :inherited-members:
--- a/docs/api/webob.txt
+++ b/docs/api/webob.txt
@@ -29,31 +29,17 @@ Accept-*
 .. autoclass:: AcceptLanguage
 
 .. autoclass:: AcceptLanguageValidHeader
-   :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
-             parse, basic_filtering, best_match, lookup, quality
-
-   .. autoinstanceattribute:: header_value
-      :annotation:
-   .. autoinstanceattribute:: parsed
-      :annotation:
+   :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
+	     __radd__, __str__, parse, basic_filtering, best_match, lookup,
+             quality
 
 .. autoclass:: AcceptLanguageNoHeader
-   :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
-             basic_filtering, best_match, lookup, quality
-
-   .. autoinstanceattribute:: header_value
-      :annotation:
-   .. autoinstanceattribute:: parsed
-      :annotation:
+   :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
+	     __radd__, __str__, basic_filtering, best_match, lookup, quality
 
 .. autoclass:: AcceptLanguageInvalidHeader
-   :members: __init__, __add__, __contains__, __iter__, __radd__, __str__,
-             basic_filtering, best_match, lookup, quality
-
-   .. autoinstanceattribute:: header_value
-      :annotation:
-   .. autoinstanceattribute:: parsed
-      :annotation:
+   :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
+	     __radd__, __str__, basic_filtering, best_match, lookup, quality
 
 .. autoclass:: MIMEAccept
    :members:
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -613,8 +613,8 @@ class AcceptLanguageValidHeader(AcceptLanguage):
 
         .. warning::
 
-           This is maintained for backward compatibility, and may be deprecated
-           in the future.
+           This is maintained for backward compatibility, and will be
+           deprecated in the future.
 
         This method was WebOb's old criteron for deciding whether a language
         tag matches a language range, used in
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -32,6 +32,17 @@ qvalue_re = r"""
 weight_re = r'[ \t]*;[ \t]*[qQ]=(' + qvalue_re + r')'
 
 
+def _item_qvalue_pair_to_header_element(pair):
+    item, qvalue = pair
+    if qvalue == 1.0:
+        element = item
+    elif qvalue == 0.0:
+        element = '{};q=0'.format(item)
+    else:
+        element = '{};q={}'.format(item, qvalue)
+    return element
+
+
 class Accept(object):
     """
     Represents a generic ``Accept-*`` style header.
@@ -468,16 +479,10 @@ class AcceptLanguageValidHeader(AcceptLanguage):
         jp;q=0.21'``.
 
         """
-        result = []
-        for range_, qvalue in self.parsed:
-            if qvalue == 1.0:
-                item = range_
-            elif qvalue == 0.0:
-                item = '{};q=0'.format(range_)
-            else:
-                item = '{};q={}'.format(range_, qvalue)
-            result.append(item)
-        return ', '.join(result)
+        return ', '.join(
+            _item_qvalue_pair_to_header_element(pair=tuple_)
+            for tuple_ in self.parsed
+        )
 
     def _old_match(self, mask, item):
         """
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -342,9 +342,7 @@ class AcceptLanguageValidHeader(AcceptLanguage):
                             ``Accept-Language`` header.
         """
         self._header_value = header_value
-
         self._parsed = list(self.parse(header_value))
-
         self._parsed_nonzero = [(m, q) for (m, q) in self.parsed if q]
 
     @property
@@ -1581,9 +1579,7 @@ class AcceptLanguageNoHeader(_AcceptLanguageInvalidOrNoHeader):
         Create an :class:`AcceptLanguageNoHeader` instance.
         """
         self._header_value = None
-
         self._parsed = None
-
         self._parsed_nonzero = None
 
     @property
@@ -1709,9 +1705,7 @@ class AcceptLanguageInvalidHeader(_AcceptLanguageInvalidOrNoHeader):
         Create an :class:`AcceptLanguageInvalidHeader` instance.
         """
         self._header_value = header_value
-
         self._parsed = None
-
         self._parsed_nonzero = None
 
     @property
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -1542,6 +1542,7 @@ class AcceptLanguageNoHeader(_AcceptLanguageInvalidOrNoHeader):
     addition operators (``+`` and ``+=``), which return a new object (see the
     docstring for :meth:`AcceptLanguageNoHeader.__add__`).
     """
+
     def __init__(self):
         """
         Create an :class:`AcceptLanguageNoHeader` instance.
@@ -1655,6 +1656,7 @@ class AcceptLanguageInvalidHeader(_AcceptLanguageInvalidOrNoHeader):
     addition operators (``+`` and ``+=``), which return a new object (see the
     docstring for :meth:`AcceptLanguageInvalidHeader.__add__`).
     """
+
     def __init__(self, header_value):
         """
         Create an :class:`AcceptLanguageInvalidHeader` instance.
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -33,6 +33,10 @@ qvalue_re = r"""
 weight_re = r'[ \t]*;[ \t]*[qQ]=(' + qvalue_re + r')'
 
 
+def _item_n_weight_re(item_re):
+    return '(' + item_re + ')(?:' + weight_re + ')?'
+
+
 def _item_qvalue_pair_to_header_element(pair):
     item, qvalue = pair
     if qvalue == 1.0:
@@ -42,6 +46,17 @@ def _item_qvalue_pair_to_header_element(pair):
     else:
         element = '{};q={}'.format(item, qvalue)
     return element
+
+
+def _list_1_or_more__compiled_re(element_re):
+    # RFC 7230 Section 7 "ABNF List Extension: #rule":
+    # 1#element => *( "," OWS ) element *( OWS "," [ OWS element ] )
+    # and RFC 7230 Errata ID: 4169
+    return re.compile(
+        r'^(?:,[ \t]*)*' + element_re +
+        r'(?:[ \t]*,(?:[ \t]*' + element_re + r')?)*$',
+        re.VERBOSE
+    )
 
 
 class Accept(object):
@@ -319,18 +334,13 @@ class AcceptLanguageValidHeader(AcceptLanguage):
         (?:-[A-Za-z0-9]{1,8})*
         )
     """
-    lang_range_n_weight_re = r'(' + lang_range_re + r')(?:' + weight_re + r')?'
+    lang_range_n_weight_re = _item_n_weight_re(item_re=lang_range_re)
     lang_range_n_weight_compiled_re = re.compile(
         lang_range_n_weight_re,
         re.VERBOSE
     )
-    # RFC 7230 Section 7 "ABNF List Extension: #rule":
-    # 1#element => *( "," OWS ) element *( OWS "," [ OWS element ] )
-    # and RFC 7230 Errata ID: 4169
-    accept_language_compiled_re = re.compile(
-        r'^(?:,[ \t]*)*' + lang_range_n_weight_re +
-        r'(?:[ \t]*,(?:[ \t]*' + lang_range_n_weight_re + r')?)*$',
-        re.VERBOSE
+    accept_language_compiled_re = _list_1_or_more__compiled_re(
+        element_re=lang_range_n_weight_re,
     )
 
     def __init__(self, header_value):
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -306,6 +306,26 @@ class AcceptLanguage(object):
     :class:`AcceptLanguageNoHeader`, and :class:`AcceptLanguageInvalidHeader`.
     """
 
+    # RFC 7231 Section 5.3.5 "Accept-Language":
+    # Accept-Language = 1#( language-range [ weight ] )
+    # language-range  =
+    #           <language-range, see [RFC4647], Section 2.1>
+    # RFC 4647 Section 2.1 "Basic Language Range":
+    # language-range   = (1*8ALPHA *("-" 1*8alphanum)) / "*"
+    # alphanum         = ALPHA / DIGIT
+    lang_range_re = (
+        r'\*|'
+        '(?:'
+        '[A-Za-z]{1,8}'
+        '(?:-[A-Za-z0-9]{1,8})*'
+        ')'
+    )
+    lang_range_n_weight_re = _item_n_weight_re(item_re=lang_range_re)
+    lang_range_n_weight_compiled_re = re.compile(lang_range_n_weight_re)
+    accept_language_compiled_re = _list_1_or_more__compiled_re(
+        element_re=lang_range_n_weight_re,
+    )
+
     @classmethod
     def _python_value_to_header_str(cls, value):
         if isinstance(value, str):
@@ -330,6 +350,33 @@ class AcceptLanguage(object):
                 header_str = str(value)
         return header_str
 
+    @classmethod
+    def parse(cls, value):
+        """
+        Parse an ``Accept-Language`` header.
+
+        :param value: (``str``) header value
+        :return: If `value` is a valid ``Accept-Language`` header, returns an
+                 iterator of (language range, quality value) tuples, as parsed
+                 from the header from left to right.
+        :raises ValueError: if `value` is an invalid header
+        """
+        # Check if header is valid
+        # Using Python stdlib's `re` module, there is currently no way to check
+        # the match *and* get all the groups using the same regex, so we have
+        # to use one regex to check the match, and another to get the groups.
+        if cls.accept_language_compiled_re.match(value) is None:
+            raise ValueError('Invalid value for an Accept-Language header.')
+        def generator(value):
+            for match in (
+                cls.lang_range_n_weight_compiled_re.finditer(value)
+            ):
+                lang_range = match.group(1)
+                qvalue = match.group(2)
+                qvalue = float(qvalue) if qvalue else 1.0
+                yield (lang_range, qvalue)
+        return generator(value=value)
+
 
 class AcceptLanguageValidHeader(AcceptLanguage):
     """
@@ -347,26 +394,6 @@ class AcceptLanguageValidHeader(AcceptLanguage):
     addition operators (``+`` and ``+=``), which return a new object (see the
     docstring for :meth:`AcceptLanguageValidHeader.__add__`).
     """
-
-    # RFC 7231 Section 5.3.5 "Accept-Language":
-    # Accept-Language = 1#( language-range [ weight ] )
-    # language-range  =
-    #           <language-range, see [RFC4647], Section 2.1>
-    # RFC 4647 Section 2.1 "Basic Language Range":
-    # language-range   = (1*8ALPHA *("-" 1*8alphanum)) / "*"
-    # alphanum         = ALPHA / DIGIT
-    lang_range_re = (
-        r'\*|'
-        '(?:'
-        '[A-Za-z]{1,8}'
-        '(?:-[A-Za-z0-9]{1,8})*'
-        ')'
-    )
-    lang_range_n_weight_re = _item_n_weight_re(item_re=lang_range_re)
-    lang_range_n_weight_compiled_re = re.compile(lang_range_n_weight_re)
-    accept_language_compiled_re = _list_1_or_more__compiled_re(
-        element_re=lang_range_n_weight_re,
-    )
 
     def __init__(self, header_value):
         """
@@ -396,33 +423,6 @@ class AcceptLanguageValidHeader(AcceptLanguage):
         A list of (language range, quality value) tuples.
         """
         return self._parsed
-
-    @classmethod
-    def parse(cls, value):
-        """
-        Parse an ``Accept-Language`` header.
-
-        :param value: (``str``) header value
-        :return: If `value` is a valid ``Accept-Language`` header, returns an
-                 iterator of (language range, quality value) tuples, as parsed
-                 from the header from left to right.
-        :raises ValueError: if `value` is an invalid header
-        """
-        # Check if header is valid
-        # Using Python stdlib's `re` module, there is currently no way to check
-        # the match *and* get all the groups using the same regex, so we have
-        # to use one regex to check the match, and another to get the groups.
-        if cls.accept_language_compiled_re.match(value) is None:
-            raise ValueError('Invalid value for an Accept-Language header.')
-        def generator(value):
-            for match in (
-                cls.lang_range_n_weight_compiled_re.finditer(value)
-            ):
-                lang_range = match.group(1)
-                qvalue = match.group(2)
-                qvalue = float(qvalue) if qvalue else 1.0
-                yield (lang_range, qvalue)
-        return generator(value=value)
 
     def __add__(self, other):
         """
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -304,6 +304,10 @@ class AcceptLanguageValidHeader(AcceptLanguage):
     7231, section 5.3.5 <7231#section-5.3.5>` to :rfc:`RFC 4647, section 2.1
     <4647#section-2.1>` to mean that only basic language ranges (and not
     extended language ranges) are expected in the ``Accept-Language`` header.
+
+    This object should not be modified. To add to the header, we can use the
+    addition operators (``+`` and ``+=``), which return a new object (see the
+    docstring for :meth:`AcceptLanguageValidHeader.__add__`).
     """
 
     # RFC 7231 Section 5.3.5 "Accept-Language":
@@ -378,6 +382,45 @@ class AcceptLanguageValidHeader(AcceptLanguage):
                     qvalue = float(qvalue) if qvalue else 1.0
                     yield (lang_range, qvalue)
             return generator(value=value)
+
+    def __add__(self, other):
+        """
+        Add to header, creating a new header object.
+
+        `other` can be:
+
+        * a ``str``
+        * a ``dict``, with language ranges as keys and qvalues as values
+        * a ``tuple`` or ``list``, of language range ``str``s or of ``tuple``
+          or ``list`` (language range, qvalue) pairs (``str``s and pairs can be
+          mixed within the ``tuple`` or ``list``)
+        * any object that returns a value for `__str__`
+        * an :class:`AcceptLanguageValidHeader`,
+          :class:`AcceptLanguageNoHeader`, or
+          :class:`AcceptLanguageInvalidHeader` instance
+
+        If neither operands are empty (header value ``''``) or
+        ``None``/:class:`AcceptLanguageNoHeader`, the two header values are
+        joined with ``', '``.
+        """
+        if isinstance(other, AcceptLanguageNoHeader):
+            return self.__class__(header_value=self.header_value)
+
+        if isinstance(other, AcceptLanguageValidHeader):
+            return create_accept_language_header(
+                header_value=self.header_value + ', ' + other.header_value,
+            )
+
+        if isinstance(other, AcceptLanguageInvalidHeader):
+            if other.header_value == '':
+                return self.__class__(header_value=self.header_value)
+            return create_accept_language_header(
+                header_value=self.header_value + ', ' + other.header_value,
+            )
+
+        return self._add_instance_and_non_accept_language_type(
+            instance=self, other=other,
+        )
 
     def __nonzero__(self):
         """
@@ -463,6 +506,16 @@ class AcceptLanguageValidHeader(AcceptLanguage):
         ):
             yield m
 
+    def __radd__(self, other):
+        """
+        Add to header, creating a new header object.
+
+        See the docstring for :meth:`AcceptLanguageValidHeader.__add__`.
+        """
+        return self._add_instance_and_non_accept_language_type(
+            instance=self, other=other, instance_on_the_right=True,
+        )
+
     def __repr__(self):
         return "{}(header_value={!r})".format(
             # ``!r`` escapes the header value
@@ -483,6 +536,43 @@ class AcceptLanguageValidHeader(AcceptLanguage):
             _item_qvalue_pair_to_header_element(pair=tuple_)
             for tuple_ in self.parsed
         )
+
+    def _add_instance_and_non_accept_language_type(
+        self, instance, other, instance_on_the_right=False,
+    ):
+        if not other:
+            return self.__class__(header_value=instance.header_value)
+
+        if isinstance(other, str):
+            other_header_value = other
+        else:
+            if hasattr(other, 'items'):
+                other = sorted(
+                    other.items(),
+                    key=lambda item: item[1],
+                    reverse=True,
+                )
+            if isinstance(other, (tuple, list)):
+                result = []
+                for element in other:
+                    if isinstance(element, (tuple, list)):
+                        element = _item_qvalue_pair_to_header_element(
+                            pair=element
+                        )
+                    result.append(element)
+                other_header_value = ', '.join(result)
+            else:
+                other_header_value = str(other)
+                if other_header_value == '':
+                    return self.__class__(header_value=instance.header_value)
+
+        new_header_value = (
+            (other_header_value + ', ' + instance.header_value)
+            if instance_on_the_right
+            else (instance.header_value + ', ' + other_header_value)
+        )
+
+        return create_accept_language_header(header_value=new_header_value)
 
     def _old_match(self, mask, item):
         """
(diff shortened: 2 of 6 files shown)


src/webob/acceptparse.py
```diff
@@ -255,11 +294,371 @@ class AcceptEncoding(Accept):
...
+        else:
+            def generator(value):
+                for match in (
+                    cls.lang_range_n_weight_compiled_re.finditer(value)
+                ):
+                    lang_range = match.group(1)
+                    qvalue = match.group(2)
+                    qvalue = float(qvalue) if qvalue else 1.0
+                    yield (lang_range, qvalue)
+            return generator(value=value)
```
mmerickel: What's the point of making this a closure?

src/webob/acceptparse.py
```diff
@@ -268,6 +667,1250 @@ def _match(self, mask, item):
...
+        if isinstance(other, AcceptLanguageNoHeader):
+            return self.__class__()
+
+        if isinstance(other, AcceptLanguageValidHeader):
+            return create_accept_language_header(
+                header_value=other.header_value
+            )
+
+        if isinstance(other, AcceptLanguageInvalidHeader):
+            return AcceptLanguageInvalidHeader(header_value=other.header_value)
```
mmerickel: I feel like adding / using an invalid header value should raise a `ValueError` versus continuing silently. Parsing an invalid header without raising is one thing... I'm on board with that. It makes it less dangerous to access the property. But continuing to use the object for anything useful seems incorrect.

src/webob/acceptparse.py
```diff
@@ -268,6 +667,1250 @@ def _match(self, mask, item):
...
+                lowered_default_tag = default_tag.lower()
+                if lowered_default_tag not in not_acceptable_ranges:
+                    return default_tag
+
+        try:
+            return default()
+        except TypeError:  # default is not a callable
+            return default
+
+    def quality(self, offer, modifier=1):
```
mmerickel: I looked in the history to see where the `modifier` was introduced and I cannot see a reason for its existence. It feels like it's probably safe to remove it. This is ambiguous with the `default_quality` parameter on the old `NilAccept` object and it also appears to be unused.

https://github.com/Pylons/webob/commit/480231cd465e56a1b67a0a9d4385fc6b0283db69#diff-078e7721c9b744590e8d0f517cc84f4dL110

src/webob/acceptparse.py
```diff
@@ -268,6 +667,1250 @@ def _match(self, mask, item):
...
+
+    def fget(request):
+        """
+        Get an object representing the header in the request.
+
+        This creates a new object (and re-parses the header) on every call.
+        """
+        return create_accept_language_header(
+            header_value=request.environ.get(ENVIRON_KEY)
+        )
```
mmerickel: I don't see a problem with re-parsing upon access. Webob historically can cache things in the environ but it's not worth it for less-popular features like accept.

src/webob/acceptparse.py
```diff
@@ -268,6 +667,1250 @@ def _match(self, mask, item):
...
+        environ, and do not involve parsing.)'
+    """
+
+    ENVIRON_KEY = 'HTTP_ACCEPT_LANGUAGE'
+
+    def fget(request):
+        """
+        Get an object representing the header in the request.
+
+        This creates a new object (and re-parses the header) on every call.
```
bertjwregeer: Remove this last sentence, the parsing shouldn't be that big of a deal. If it is something that ends up dominating a lot of time for users we can figure out how to speed it up at that time. This exposes how it is implemented rather than what it does.

bertjwregeer:
1. module scope would be fine, leaving it as is would be fine too. I don't mind the lot of noise, so long as it causes a failure and the test suite is not considered to have passed.
2. No hard or fast preferences. pytest prefers using test functions instead of test classes.
3. Since this is not in the hot path, I am not worried about reparsing. If it becomes an issue or is significantly slow enough to show up in profiles, let's not worry about it for now.
4. pytest will print them, downstream users should be running with warnings turned on IMHO especially if they are upgrading. I am not too worried about it, we write good docs for changes.
5. Let's deprecate them, and remove em 2+ versions from now. I am less worried about maintaining something that gives you the wrong answer because it probably is not what people are expecting...
6. Don't mark it as an ABC, we don't have anything else in WebOb that is marked explicitly as a base class. I'd like to have @mmerickel's opinion on this though
7. I need to re-read your comments and then I'll be able to answer this!

I've reviewed it, and will continue to do so over the next couple of days. This PR is looking great, keep up the good work!

src/webob/acceptparse.py
whiteroses: It is so that the `ValueError` is raised when `.parse()` is called, rather than when `.next()`/`.__next__()` is first called on the iterator. So if it's not a closure, we would have to attempt to consume the first item from the iterator before we know whether the header is invalid. But it could work either way, and depends on when you want the `ValueError` to be raised.

(The `else:` is unnecessary &mdash; I'll remove it.)

src/webob/acceptparse.py
whiteroses: So: deprecation warnings, remove at later date?

I think the `modifier` might have been intended to parallel the [`server_quality`](https://github.com/Pylons/webob/blob/master/src/webob/acceptparse.py#L131) in `Accept.best_match()`, but it wasn't spotted that the `modifier` only needs to be multiplied to `bestq` at the end. So we may want to keep it around for backward compatibility until `.best_match()` can be deprecated? (`default_quality` definitely doesn't do anything though.)

Can I ask how you tracked down which commit added the `modifier`?

src/webob/acceptparse.py
whiteroses: To be honest, I don't know why we would want to add headers? (I mean genuinely don't know, would like to learn.) I just thought that if adding headers was a way to build up a valid header from parts, then we could build a valid header from `AcceptLanguageNoHeader`, `AcceptLanguageInvalidHeader`, and `AcceptLanguageValidHeader` parts (say, `None` + `, ` + `en`).

src/webob/acceptparse.py
whiteroses: Should I remove the paragraph about parsing from `doc` as well?

whiteroses:
1\.  I imagine it'd be best to stick to one style for WebOb, and as the imports for the existing tests in acceptparse.py are module scope and you're happy with it, I'll switch to module scope too.

2\.  I mentioned a couple of advantages for using test classes, and don't know the advantages of using test functions, so would like to know what they are, as it does seem to be the style pytest prefers (but don't think I've ever come across any explanations as to why). Either way, could you give me a ruling on whether I should switch? :)

5\.  Sorry, just to be clear, that's a no on fixing them?

7\.  There's now a discussion on this [here](https://github.com/Pylons/webob/pull/335#discussion_r130519570).

Thanks for reviewing!

src/webob/acceptparse.py
mmerickel: I can't find any cases where the `modifier` is actually used internally and it isn't documented. In my mind we could just remove it but if we are worried we could deprecate it first.

> Can I ask how you tracked down which commit added the modifier?

I just used blame to traverse the history. Sometimes you need to blame multiple times anchored at commits as you go backward but in this case it was the last change to that line of code in the master branch's history.

src/webob/acceptparse.py
mmerickel: Well I think the common case here would be to construct a list of languages from strings. For example, `reseponse.accept_language += 'en-US;q=0.8'`. Appending actual `AcceptLanguage*` objects is probably more rare.

mmerickel:
> 5. Sorry, just to be clear, that's a no on fixing them?

I believe he's saying to switch from `PendingDeprecationWarning` to `DeprecationWarning` and just get on with it. That's fine with me too but we'll need to coordinate that with openstack/pyramid who are using `best_match` right now. I would argue to just leave it at pending and let us sort it out later.

mmerickel:
> 6. Don't mark it as an ABC, we don't have anything else in WebOb that is marked explicitly as a base class. I'd like to have @mmerickel's opinion on this though

I don't think there's any wins here by marking it as a base class. It is not intended to be extended by third party developers and is instead an internal hierarchy where most people should only care about the top-level `AcceptLanguage` API.

src/webob/acceptparse.py
mmerickel: Thanks for pointing this out - I forgot about that behavior with generators!

src/webob/acceptparse.py
```diff
@@ -376,8 +376,8 @@ def parse(cls, value):
 
         :param value: (``str``) header value
         :return: If `value` is a valid ``Accept-Language`` header, returns a
-                 generator that yields (language range, quality value) tuples,
-                 as parsed from the header from left to right.
+                 generator iterator that yields (language range, quality value)
```
mmerickel: just "iterator" would be appropriate terminology here. The fact that it is a generator is not useful to the user.

src/webob/acceptparse.py
whiteroses: Oh, it's right there on the first `git blame` &mdash; I must've confused it with something else I was looking for, where `git blame` only showed the code being moved. Didn't know about anchoring at commits, will have to `man git blame` &mdash; that's good to know, thanks!

src/webob/acceptparse.py
whiteroses: Yeah, I can't think of a case where we would want to add actual header objects. But I see the difference between adding strings where each string would be an invalid header individually, and adding an actual `AcceptLanguageInvalidHeader`. Ok, will change.

whiteroses: Should adding an `AcceptLanguageInvalidHeader` raise a `ValueError` or a `TypeError`?

mmerickel: It would be a `TypeError` here but like I said above I no longer think raising is the right answer here.

bertjwregeer:
Don't switch from test classes to test functions. The reason they are preferred by pytest is that instead of having a setup function in the class, you just have a fixture that the test relies on. This way if the test doesn't explicitly need that fixture then it can leave it off (and probably should).

The pytest way is not used within the webob project at the moment, so don't worry about it too much.

src/webob/acceptparse.py
whiteroses: Ah sorry, I'd missed that comment somehow. Ignoring invalid headers makes sense to me, especially with your scenario of `request.accept_language += 'en-US;q=0.8'`. I think it depends on what we want addition to mean: my implementation in the PR is adding two header values, whatever they are, and deciding afterwards from the result what class of header it is. Your suggestion is saying, before the addition, that an invalid header is not a thing that should exist; and if we must add an invalid header, then we'll ignore it rather than raise. So the current implementation in the PR is adding them as strings, then reacting to the result; your suggestion would be adding them as individual *headers*, where an invalid header should be ignored. Both make sense in their own way, and it depends on which meaning you want `__add__`/`__radd__` to represent.

If adding an invalid header (as an object) is to be ignored, that means adding an invalid header *value* (as a str, dict, etc.) should also be ignored, right?

whiteroses:
Started changing imports to module-scope, but it just occurred to me: if we use module-scope imports instead of `self._get_function()` etc., then if we change the name of the function, we would need to update many tests. Is that okay?

bertjwregeer:
Yep. Mass search and replace is just fine. I am not that worried about it.

src/webob/acceptparse.py
whiteroses: Understood and fixed.

src/webob/acceptparse.py
whiteroses: @mmerickel Were you waiting to see what @bertjwregeer and @stevepiercy think here, or should I go ahead and remove/deprecate?

src/webob/acceptparse.py
whiteroses: @mmerickel Not sure if you've had time to consider this &mdash; not doing anything with the `InvalidHeader` is probably the safer option, so I can make the change if you're happy with the decision?

whiteroses:
@bertjwregeer Is [this](https://github.com/Pylons/webob/pull/335#issuecomment-319396172) what you meant/should I change them to `DeprecationWarning`s?

bertjwregeer:
Yeah, instead of PendingDeprecation just use Deprecation. The OpenStack guys are generally pretty quick about fixing their software, and we are only making it better for them!

whiteroses:
@mmerickel said [here](https://github.com/Pylons/webob/pull/335#issuecomment-319396172) that "we'll need to coordinate that with openstack/pyramid who are using best_match right now. I would argue to just leave it at pending and let us sort it out later." Maybe we can leave it on PendingDeprecation for `Accept-Language` for now, and once I finish the other three headers, provided all is well and you agree we are ready, switch all of them to Deprecation together? (I think Pyramid only uses `best_match()` for the `Accept` header, though I'd have to check. I'm also hoping to have time in the last week or so of August to look at Pyramid's issues with the `Accept` header, where I may be able to update the use of `best_match()` for you.)

src/webob/acceptparse.py
mmerickel: Remove it.

src/webob/acceptparse.py
mmerickel: I think ignoring it is fine (anything wins / replaces it).

src/webob/acceptparse.py
whiteroses: I've removed `default_quality`. I don't think we can remove `modifier`, as it is [documented](https://webob.readthedocs.io/en/stable/api/webob.html), is actually used within the method and affects the return value. As `.quality()` will be deprecated, I think we can leave it?

src/webob/acceptparse.py
whiteroses: I've made the changes: `InvalidHeader`s and invalid header values are now treated like `NoHeader`s when `__add__`- and `__radd__`ing.

src/webob/acceptparse.py
```diff
@@ -255,11 +294,382 @@ class AcceptEncoding(Accept):
...
-    Represents an ``Accept-Language`` header.
+    Represent an ``Accept-Language`` header.
+
+    Base class for :class:`AcceptLanguageValidHeader`,
+    :class:`AcceptLanguageNoHeader`, and :class:`AcceptLanguageInvalidHeader`.
     """
-    def _match(self, mask, item):
+
+    @classmethod
+    def python_value_to_header_str(cls, value):
```
mmerickel: How about naming this `format`?

src/webob/acceptparse.py
mmerickel: I would still just remove it but whatever... you could just deprecate it and play nice.

src/webob/acceptparse.py
```diff
@@ -255,11 +294,382 @@ class AcceptEncoding(Accept):
...
+    def parsed(self):
+        """
+        (``list``) Parsed form of the header.
+
+        A list of (language range, quality value) tuples.
+        """
+        return self._parsed
+
+    @classmethod
+    def parse(cls, value):
```
mmerickel: Should this be on `AcceptLanguage` instead of `AcceptLanguageValidHeader`? It seems to me that in general the subclasses should be hidden as an implementation detail and public apis like this should be on the base class.

src/webob/acceptparse.py
mmerickel: It feels like maybe method and `parse` are a little ambiguous. It's semi-feeling like there should just be one api that takes a python value and returns an accept language header. From there you can use `.header_value` to get the actual header string.

```python
header = AcceptLanguage.parse(...) # possibly more clear than the current `create_accept_language_header`
items = AcceptLanguage.parse(...).items # basically the same as `.parsed` right now
header_value = AcceptLanguage.parse(...).value # equivalent to `python_value_to_header_str`
```

src/webob/acceptparse.py
whiteroses: I meant it's already being deprecated with the rest of the `.quality()` method, so it seems not worth even the small risk of breaking anything &mdash; but if you want it gone earlier, I can remove it.

src/webob/acceptparse.py
whiteroses: But `.parse()` is documented api that (I assume) we can't break? It currently returns an iterator, so I have kept that the same in my PR.

In case it's not clear, `.python_value_to_header_str()` is not intended to be public API &mdash; it's not documented, and is just a helper method. I only didn't prefix the name with an underscore because it needs to be called from `accept_language_property` too (but maybe I should?) So there's only one API to create a header from a non-`str` Python value: `AcceptNoHeader() + ...`.

src/webob/acceptparse.py
whiteroses: I would've put it on `AcceptLanguage`, but `AcceptLanguageNoHeader` and `AcceptLanguageInvalidHeader` would then inherit `.parse`, and it doesn't seem to make sense for a `NoHeader` or an `InvalidHeader` to have a `.parse`. It seems to make more sense for a `ValidHeader` to attempt to `.parse`, and raise an exception if that fails.

And in this design, the subclasses are not really just implementation detail any more. For example, we can check whether a header is an invalid or no header with e.g. `isinstance(header_object, AcceptLanguageInvalidHeader)`, and that is by design. I definitely started off wanting to hide everything behind the `AcceptLanguage` class, but it was getting too complex, and the functionality seems to be much clearer as objects of three distinct classes. (Right now the `Accept-Language` header in WebOb is two classes, `AcceptLanguage` and `NilAccept`, which are not even subclasses of the same base class, and have no connection with each other.)

And while `ValidHeader`, `InvalidHeader` and `NoHeader` have many common methods, it makes sense that they do not necessarily share the exact same interface &mdash; there would be things we can do with a `ValidHeader` that we wouldn't do with e.g. a `NoHeader`. I made them inherit from `AcceptHeader` mainly so they all i.d. as `AcceptHeader`, but you can see that `AcceptHeader` otherwise has very little in it. I would have to spend more time diving back into it all to remember all the reasons why I stopped trying to put everything behind the `AcceptLanguage` class (as I've moved on to working on the `Accept` header), but I think my reasoning was along those lines.

src/webob/acceptparse.py
```diff
@@ -686,10 +686,16 @@ def basic_filtering(self, language_tags):
...
         :rfc:`RFC 4647, section 3.3.1 <4647#section-3.3.1>`. It filters the
-        tags in the `language_tags` argument, and returns a list of the tags
-        that match the language ranges in the header according to the Basic
-        Filtering matching scheme, in descending order of preference, together
-        with the qvalue of the range each tag matched.
+        tags in the `language_tags` argument and returns the ones that match
+        the header according to the matching scheme. The returned list is a
+        list of (language tag, qvalue) tuples, in descending order of qvalue;
+        if two or more tags have the same qvalue, they are returned in the same
+        order as the order in the header of the ranges they matched. If the
```
stevepiercy: This is less klunky: "order as that in the header"

src/webob/acceptparse.py
```diff
@@ -686,10 +686,16 @@ def basic_filtering(self, language_tags):
...
         :rfc:`RFC 7231, section 5.3.5 <7231#section-5.3.5>`, and defined in
         :rfc:`RFC 4647, section 3.3.1 <4647#section-3.3.1>`. It filters the
-        tags in the `language_tags` argument, and returns a list of the tags
-        that match the language ranges in the header according to the Basic
-        Filtering matching scheme, in descending order of preference, together
-        with the qvalue of the range each tag matched.
+        tags in the `language_tags` argument and returns the ones that match
+        the header according to the matching scheme. The returned list is a
+        list of (language tag, qvalue) tuples, in descending order of qvalue;
+        if two or more tags have the same qvalue, they are returned in the same
```
stevepiercy: Let's break this into two sentences to be consistent with the next sentence. IOW, "...qvalue. If..."

src/webob/acceptparse.py
```diff
@@ -686,10 +686,16 @@ def basic_filtering(self, language_tags):
...
+        order as the order in the header of the ranges they matched. If the
+        matched range is the same for two or more tags (i.e. their matched
+        ranges have the same qvalue and the same position in the header), then
+        they are returned in the same order as their order in the
```
stevepiercy: Shorten: "returned in the same order as that in the"

src/webob/acceptparse.py
```diff
@@ -686,10 +686,16 @@ def basic_filtering(self, language_tags):
...
+        `language_tags` argument. (If `language_tags` is unordered, e.g. if it
+        is a set or a dict, then that order may not be reliable.)
```
stevepiercy: Remove parentheses, as this is really not a parenthetical statement, but a sentence of distinct importance.

src/webob/acceptparse.py
```diff
@@ -701,29 +707,18 @@ def basic_filtering(self, language_tags):
...
-        The method returns a list of tuples of the form (language tag, qvalue),
-        in descending order of preference: in descending order of qvalue, and
-        if two tags have equal qvalues, we consider the tag whose matched range
-        appears earlier in the header to have higher preference. If the matched
-        range is the same for two or more tags (i.e. their matched ranges have
-        the same qvalue and the same position in the header), their order in
-        the `language_tags` argument is used as tiebreaker. (If `language_tags`
-        is unordered, e.g. if it is a set or a dict, then that order may not be
-        reliable.)
+           and there is one or more ``*`` language range in the header, then:
```
stevepiercy: strip colon "then"

src/webob/acceptparse.py
```diff
@@ -701,29 +707,18 @@ def basic_filtering(self, language_tags):
...
+           if any of the ``*`` language ranges have ``q=0``, the language tag
+           is filtered out. Otherwise, the language tag is considered a match.
```
stevepiercy: "is filtered out, else the language tag..."

src/webob/acceptparse.py
mmerickel: > And in this design, the subclasses are not really just implementation detail any more. For example, we can check whether a header is an invalid or no header with e.g. isinstance(header_object, AcceptLanguageInvalidHeader), and that is by design. I definitely started off wanting to hide everything behind the AcceptLanguage class, but it was getting too complex, and the functionality seems to be much clearer as objects of three distinct classes. (Right now the Accept-Language header in WebOb is two classes, AcceptLanguage and NilAccept, which are not even subclasses of the same base class, and have no connection with each other.)

Checking `isinstance` is considered to be poor form in most any subclassing hierarchy and is a hacky escape hatch. If you know up front that an `AcceptLanguage` header can be in one of these 3 states then the parent class should expose enough information to determine that without having to check the type of class you're getting. I really do think that these classes should be implementation details just like the old design of `AcceptLanguage` and `NilAccept`. The primary interface should be a function that parses strings (or other python objects like `python_object_to_header`) and returns *something* implementing the `AcceptLanguage` api. The fact that may return different classes is just an implementation convenience and the api design should not be dependent on it.

src/webob/acceptparse.py
```diff
@@ -686,10 +686,16 @@ def basic_filtering(self, language_tags):
         suggested as a matching scheme for the ``Accept-Language`` header in
         :rfc:`RFC 7231, section 5.3.5 <7231#section-5.3.5>`, and defined in
         :rfc:`RFC 4647, section 3.3.1 <4647#section-3.3.1>`. It filters the
-        tags in the `language_tags` argument, and returns a list of the tags
-        that match the language ranges in the header according to the Basic
-        Filtering matching scheme, in descending order of preference, together
-        with the qvalue of the range each tag matched.
+        tags in the `language_tags` argument and returns the ones that match
```
whiteroses: I think I looked this up at the time and found in [PEP 287](https://www.python.org/dev/peps/pep-0287/):

> Text enclosed in single backquotes is recognized as "interpreted text", whose interpretation is application-dependent. In the context of a Python docstring, the default interpretation of interpreted text is as Python identifiers. The text will be marked up with a hyperlink connected to the documentation for the identifier given. 

There is an example following that paragraph that shows the use of single backquotes for method parameters. And earlier in the PEP, it mentions the use for double backquotes, but for "program I/O or code snippets":

> Inline literals use double-backquotes to indicate program I/O or code snippets. No markup interpretation (including backslash-escape [\] interpretation) is done within inline literals.

Does that matter, or should I go with double backquotes?

whiteroses:
@stevepiercy Other than the one on backquotes that I had a question on, I've made the changes you requested. Thanks!

src/webob/acceptparse.py
stevepiercy: I didn't know about that. Let's leave it as single backticks.

As an aside, I think I might have screwed up the markup in Pyramid docstrings along the way.

src/webob/acceptparse.py
whiteroses: I definitely understand seeing `AcceptLanguageValidHeader`, `AcceptLanguageNoHeader` and `AcceptLanguageInvalidHeader` as implementation detail and wanting to hide them behind an `AcceptLanguage` class &mdash; as mentioned before, that was very much what I had in mind to start with, and I ran into enough obstacles that I decided to switch away from that approach. It may be that it is possible and preferable, though it may take some time to make the change.

When you say "returns *something* implementing the `AcceptLanguage` api", do you mean return an instance of one of the three classes, or an `AcceptLanguage` instance that, say, proxies any attribute access and method calls to an instance it holds of one of the three subclasses?

Are we then to combine the docstrings for the three cases for each method? While there is some overlap, it makes the docstrings (many of which are already quite long) even longer, and imo makes them a bit more difficult to understand. I think that was a large part of what turned me away from that approach: that it bounds three very distinct cases together, when each case (especially the ValidHeader case) is often complicated enough by itself already. Keeping them separate on the interface level seems to me to make them easier to understand.

If using `isinstance` is a problem, should there be an `is_valid` attribute? (What should `is_valid` return for a NoHeader? It's not really valid *or* invalid &mdash; there is no header at all.) What should be the attribute to indicate that the header is a NoHeader, or should we leave the user to check that `.header_value` is `None`?

whiteroses: I forgot to ask: where you said 'these classes should be implementation details just like the old design of `AcceptLanguage` and `NilAccept`' &mdash; isn't `NilAccept` very much an implementation detail? It was shared across three headers for convenience of implementation, and not hidden behind `AcceptLanguage` at all?

src/webob/acceptparse.py
whiteroses: I was going to say that I didn't think Sphinx did anything special with single backticks other than italics, but I just checked and, there looks to be something potentially useful in the note at the top of [this page](http://www.sphinx-doc.org/en/stable/markup/inline.html).

stevepiercy: Single backticks alone render in HTML markup as `<cite>`:

```html
<cite>language_tags</cite>
```

As an aside, I wish Sphinx had used `<code>` or `<pre>` instead, because `<cite>` is semantically incorrect. `¯\_(ツ)_/¯`

By default most web browsers use italics for styling `<cite>`. We could style the `<cite>` tags however we want, say as preformatted or monospace code, but italics are fine for now.

The page you linked uses a `:role:` preceding the single backticked item. You could try the `:any:` role preceding `language_tags` and see if magic happens or if the HTML markup changes to something not italicized. Else there might be a specific existing role under the [Python domain in Sphinx](http://www.sphinx-doc.org/en/stable/domains.html#the-python-domain). Finally we could define a custom role for parameters, if none of the existing ones are satisfactory, but I think that would be too much effort for very little or no benefit.

src/webob/acceptparse.py
mmerickel: > When you say "returns something implementing the AcceptLanguage api", do you mean return an instance of one of the three classes, or an AcceptLanguage instance that, say, proxies any attribute access and method calls to an instance it holds of one of the three subclasses?

It would return one of 3 different instances that conform to the same api.

> If using isinstance is a problem, should there be an is_valid attribute? (What should is_valid return for a NoHeader? It's not really valid or invalid — there is no header at all.) What should be the attribute to indicate that the header is a NoHeader, or should we leave the user to check that .header_value is None?

Yes I think `header_value` might be a good way to inspect things but you could also just define some `status` property that is `valid` or `invalid` or `missing` if anyone really needs to know.

> isn't NilAccept very much an implementation detail

Yes, it was an implementation detail - and I hope we can do something similar here.

As far as docstrings go you can just inherit the `AcceptLanguage` docstrings if you want or you can define them separately for each subclass. In general the user is expected to program against the contract defined in the `AcceptLanguage` class - not the specifics of each subclass.

src/webob/acceptparse.py
bertjwregeer: Add an underscore to the name. This way it is markedly non-public and makes it simpler to see so when reading the code.

src/webob/acceptparse.py
bertjwregeer: `AcceptLanguageValidHeader` would basically move wholesale into `AcceptLanguage` if I understand @mmerickel's comments correctly, and then the sub-classes that are `NoHeader` or `InvalidHeader` would have `.parse()` and others `raise ValueError` or some other exception? 

Do I understand this right @mmerickel?

src/webob/acceptparse.py
```diff
@@ -268,6 +678,1197 @@ def _match(self, mask, item):
...
+              >>> header.best_match(offers=['zh-Hans'])
+              >>> header.best_match(offers=['zh-Hans']) is None
+              True
+
+        """
+        warnings.warn(
+            'The behavior of AcceptLanguageValidHeader.best_match is '
+            'currently being maintained for backward compatibility, but it may'
+            ' be deprecated in the future as it does not conform to the RFC.',
+            PendingDeprecationWarning,
```
bertjwregeer: Just make this a `DeprecationWarning`.

src/webob/acceptparse.py
```diff
@@ -268,6 +678,1197 @@ def _match(self, mask, item):
...
+          during the subtag truncation search for `default_range`, that are an
+          exact match for a non-``*`` range with ``q=0`` in the header, are
+          considered not acceptable and ruled out.
+
+        - If there is a ``*;q=0`` in the header, then `default_range` and
+          `default_tag` have no effect, as ``*;q=0`` means that all languages
+          not already matched by other ranges within the header are
+          unacceptable.
+        """
+        assert not (default_tag is None and default is None), \
```
bertjwregeer: Instead of using assert, use an if statement. assert statements may be turned off globally, and while I tend to believe that if you break it you've bought it, I'd prefer not to have asserts in anything put testing code.

Unless it is a condition that is not likely to ever happen, and is just a sanity check.

src/webob/acceptparse.py
```diff
@@ -268,6 +678,1197 @@ def _match(self, mask, item):
...
+            '`default_tag` and `default` arguments cannot both be None.'
+
+        # We need separate `default_tag` and `default` arguments because if we
+        # only had the `default` argument, there would be no way to tell
+        # whether a str is a language tag (in which case we have to check
+        # whether it has been specified as not acceptable with a q=0 range in
+        # the header) or not (in which case we can just return the value).
+
+        assert default_range != '*', 'default_range cannot be *.'
```
bertjwregeer: Same thing here, no assert please.

src/webob/acceptparse.py
bertjwregeer: > Return the quality of the given offer. Returns None if there is no match (not 0).

That's the docs on `.quality()` at the moment. Feel free to remove `modifier` since it is not documented what it does, or why it is set to 1 by default.

Set it to 1.

src/webob/acceptparse.py
```diff
@@ -268,6 +678,1197 @@ def _match(self, mask, item):
...
+              >>> header.quality(offer='zh-Hans')
+              >>> header.quality(offer='zh-Hans') is None
+              True
+
+        """
+        warnings.warn(
+            'The behavior of AcceptLanguageValidHeader.quality is'
+            'currently being maintained for backward compatibility, but it may'
+            ' be deprecated in the future as it does not conform to the RFC.',
+            PendingDeprecationWarning,
```
bertjwregeer: Make this a `DeprecationWarning`. I really don't think it is a good idea to keep this around for very long.

src/webob/acceptparse.py
```diff
@@ -268,6 +678,1197 @@ def _match(self, mask, item):
...
+        representing the header is created every time we *get* the value of the
+        property. (*set* and *del* change the header value in the request
+        environ, and do not involve parsing.)'
+    """
+
+    ENVIRON_KEY = 'HTTP_ACCEPT_LANGUAGE'
+
+    def fget(request):
+        """Get an object representing the header in the request."""
+        return create_accept_language_header(
```
bertjwregeer: Here we return a brand new `AcceptLanguage*` object each time, yet when we assign a new `AcceptLanguage` using fset we set it directly to the AcceptLanguage object.

Would it make sense here to check if it is a string, and if so then run the `create_accept_language_header` function, followed by setting the environ key to the newly created object? This way if a user calls `request.accept_language` twice they only have the overhead of creating the object the first time? If they modify `request.environ` directly it would no longer be an `AcceptLanguage` object and would correctly re-create it.

----

Another thing I noticed is that with your new `__add__` functions since they don't modify the existing one but create a new one something like this is not possible, or am I mistaken:

```
request.accept_langauge += 'nl-NL;q=0.7'
```

Since that would return a new object and `request.accept_language` would be untouched. I can live with that... just not what I expect at the moment based upon the other properties.

bertjwregeer:
With these changes, I hope that we have satisfactorily fixed the issue brought up by the original poster in https://github.com/Pylons/webob/issues/256

src/webob/acceptparse.py
mmerickel: My point is that the subclasses shouldn't have any public apis that aren't defined on the `AcceptLanguage` class as no one should care which subclass they are actually using. It's standard duck typing of an api. You get back some class that conforms to the `AcceptLanguage` interface after parsing a value.

bertjwregeer: :+1: I agree with this.

src/webob/acceptparse.py
whiteroses: It doesn't assign a new `AcceptLanguage` when we fset though: fset only changes the header value in the environ, so that the next time we fget, a brand new `AcceptLanguage*` object is created using that new header value in the environ. This is the same as how it current works in `accept_property`.

I can't quite understand the second paragraph? In terms of overhead, we talked a while back about whether we should cache the object, as a new object was being created on every attribute read, and you said that it wasn't a concern at that point &mdash; is that related to what you're asking here?

With the `__add__` functions,

    request.accept_language += 'nl-NL;q=0.7'

is the same as

    request.accept_language = request.accept_language + 'nl-NL;q=0.7'

so it works :)

whiteroses:
@bertjwregeer: I've made the changes you requested &mdash; thanks for reviewing! With #256, `.lookup()` covers that specific case; I can make a comment on the issue explaining this after the PR is merged. I'd also like to update some of the information in these issues, as my understanding of them has changed, and the information there may be useful to others (I was already finding these issues turning up near the top in search results when looking up various aspects of Accept* headers implementation!) We can still comment on closed issues, right?

bertjwregeer:
Yes, you may definitely comment on closed issues!

Were you able to make the changes that @mmerickel suggested regarding the duck typing? I think that is important before merging this in!

src/webob/acceptparse.py
whiteroses: Can I ask what changes would you actually like to see? You may remember that I asked if we should use an abstract base class to enforce the interface, so that was what I was trying to do. The api of the three classes are already very similar. But I am truly not understanding what you are asking for. I need most to understand what @mmerickel  meant by

> 'these classes should be implementation details just like the old design of AcceptLanguage and NilAccept'

and 

> > isn't NilAccept very much an implementation detail

> Yes, it was an implementation detail - and I hope we can do something similar here.

Because it is working the same now as it did before, only with different classes: it was `AcceptLanguage` and `NilAccept` before, and `AcceptLanguageValidHeader`, `AcceptLanguageNoHeader`, and `AcceptLanguageInvalidHeader` now. `NilAccept` is not hidden within `AcceptLanguage`, and there is no requirement I can see that `AcceptLanguage` and `NilAccept` have the same api (for example, `NilAccept` does not have `.parse`. And `.quality` had different parameters, which we just removed.)

If we go into the Python interpreter and look at the object we get with `request.accept_language`, depending on the header value we supply, we would get either an `AcceptLanguage` or a `NilAccept`: `NilAccept` is not hidden. If I wanted to know what attributes and methods were available when I saw the `NilAccept`, I would look up the documentation for `NilAccept`, not `AcceptLanguage`.

If you'd like the three classes to have the exact same API, I can make that happen. But they would still identify as the individual subclasses, like they are in the PR now, and like they were before this PR.

Do you want this:

> It would return one of 3 different instances that conform to the same api.

Which is basically what it is now, but making sure the api is exactly the same across the three subclasses, and adding an attribute that identify the object as being in one of the three states instead of using `isinstance()`?
 
Or this:

> AcceptLanguageValidHeader would basically move wholesale into AcceptLanguage if I understand @mmerickel's comments correctly, and then the sub-classes that are NoHeader or InvalidHeader would have .parse() and others raise ValueError or some other exception?

Or, as I mentioned earlier, hide all three classes within an `AcceptLanguage` class by, say, proxying attribute and method access? (Which is what I think of as truly hiding the implementation detail: the object would identify as `AcceptLanguage` at all times.)

Please, help me understand the change you would like to see.

whiteroses: I re-read this:

> You get back some class that conforms to the AcceptLanguage interface after parsing a value.

And if I am understanding correctly, the classes in the PR already do this &mdash; I already made their interface the same, other than `.parse`. The only issue is we need an attribute to identify which of the three states the instance is in, instead of relying on `isinstance()`. Do I have that right? Is that the only change you are looking for?

bertjwregeer: - `AcceptLanguageValidHeader`
- `AcceptLanguageNoHeader`
- `AcceptLanguageInvalidHeader`

Should all have the same functions available, even if in the latter two cases (`AcceptLanguageNoHeader` and `AcceptLanguageInvalidHeader`) the functions don't do anything other than raise a `ValueError`.

This way no matter what class you got, you could for example call `.parse()` on it.

What @mmerickel is saying, and now that I've re-read what he meant is: move `.parse()` in it's entirety from `AcceptLanguageValidHeader` to `AcceptLanguage`.

No further code changes are necessary.

bertjwregeer: Actually, you'll also need to move https://github.com/Pylons/webob/pull/335/files#diff-5b0e7ad4b74e3dd8fb49c426d65a5115R351 (line 351 through 369) to `AcceptLanguage`.

src/webob/acceptparse.py
whiteroses: I think the idea of the single backticks in Sphinx and the `default_role` is that we wouldn't need to specify a role &mdash; there's an example [here](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-any), from one of the links in the note. So if they are cross-references, `<cite>` kind of makes sense? But I don't think that works for parameters like `language_tags`, so it doesn't help us there.

Italics kinda makes sense to me for parameters, although `<cite>` doesn't without cross-references. If you'd prefer them styled differently, maybe we can look into it after GSoC?

src/webob/acceptparse.py
bertjwregeer: Excellent. No worries, not sure what I was thinking. We are good here!

src/webob/acceptparse.py
stevepiercy: I think we're pretty much in agreement. If doing this:

```rst
:any:`language_tags`
```

...does not generate a useful link or reference, then there's no point in pursuing it.

As far as styles go, I've opened up a [new issue to resolve this part of the discussion](https://github.com/Pylons/pylons-sphinx-themes/issues/7).

src/webob/acceptparse.py
whiteroses: I've moved `.parse` and the regexes as you requested. Should I add something like a `status` property on the subclasses as well, as @mmerickel suggested [here](https://github.com/Pylons/webob/pull/335#discussion_r134620078)? Right now the only way to easily tell whether a header object is valid, invalid or None is to use `isinstance`, which @mmerickel said was considered to be poor form. Maybe call it `state`?

whiteroses:
@bertjwregeer Have made the changes, please see [here](https://github.com/Pylons/webob/pull/335#discussion_r135152443).

src/webob/acceptparse.py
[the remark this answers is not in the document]
whiteroses: Ah, I see what you mean.

I tried

    :any:`language_tags`

and got an InvocationError: `'any' reference target not found: language_tags`. Skimming through the docs, it doesn't look like there's currently a way to cross-reference a parameter. I'll come back to your issue after the work on the headers is done and merged, to see if I can help (if it's not resolved by then). Thanks.

src/webob/acceptparse.py
bertjwregeer: I am not opposed to it, but at the same time I am not sure it is useful. If you see a good argument for it, then please do so, but as a user of the library I care more that `request.accept_language` allows me to call some functions on it like `.lookup` or `.basic_filtering` on it. So long as those functions return something useful then I as a user am happy.

I don't know that previously we had a way other than using `isinstance()` to verify what state an Accept header was in...

mmerickel: I agree that I do not see a strong need for exposing this information but if we wanted to then probably status would be a good way to do it.

whiteroses: I guess there's no obvious reason why we'd need to react to or log an invalid header, and the property could always be easily added if needed. Ok, let's leave it?

docs/api/webob.txt
```diff
@@ -16,9 +16,25 @@ Accept-*
...
+
+.. autoclass:: AcceptLanguageNoHeader
+   :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
+	     __radd__, __str__, parse, basic_filtering, best_match, lookup,
+             quality
+
+.. autoclass:: AcceptLanguageInvalidHeader
+   :members: header_value, parsed, __init__, __add__, __contains__, __iter__,
+	     __radd__, __str__, parse, basic_filtering, best_match, lookup,
+             quality
```
mmerickel: Can we remove the subclasses from the public api?

whiteroses: We can if we combine the docstrings for all the subclasses for each method and put them all into `AcceptLanguage`. This would imo make some already complicated docstrings even more complicated, adding another layer to the branching. And as I mentioned before, the instances would still i.d. as the subclasses (I guess we could change the `__repr__` to indicate that it is a subclass of `AcceptLanguage`, so people would know to look there?) If you want me to go ahead and make the change, I will &mdash; it's not trivial, and will take some time and thought on how to combine the docstrings from the three subclasses for each method. (I would also have to do the same for the other three headers.)

Would it be possible to merge this so I can push changes for this and the other three headers before the deadline tomorrow to a more appropriately-named PR, or should I push the changes for all four headers to this one?

mmerickel: It's fine if we want to merge this now. I'm happy with that.

Part of the idea of making the subclasses "hidden" is not that people don't know about them but that they don't care. I'm more worried about the docs the user sees than the actual objects they see in the repl. With that in mind we'd want to make the docstrings focus on valid headers with just a sentence or two in each that explains what happens with invalid/missing data. I can't imagine that will make the docstrings that much more complex or harder to parse but I could be wrong!

whiteroses: I was worrying whether it would be clear enough that people should look at the docs for `AcceptLanguage` if the objects identify as the subclasses, but I understand what you are looking for, and will try to combine the docstrings and document one api under the base class after I've finished wrapping up the four headers and all the necessary submissions for gsoc. Could you please merge this PR if everything is ok, so I can open a new PR and push the other changes? (Sorry about the last-minute commits, they are minor changes to fix small mistakes.) Thanks!

mmerickel: Thanks again @whiteroses I'm happy with all of this.

bertjwregeer: Alright, any future work can be done in a different PR.

bertjwregeer:
Thank you very much @whiteroses, any further changes can be made on a new PR against master!

whiteroses:
Thanks @bertjwregeer!

PR #338, opened 2017-08-29
Update and fix Accept* headers handling

whiteroses:
This PR is for the work on the `Accept`, `Accept-Charset` and `Accept-Encoding` headers, and any further work on the `Accept-Language` header, following on from [this merged PR](https://github.com/Pylons/webob/pull/335).


bertjwregeer:
I haven't forgotten about this! Will take a look tomorrow. Thanks @whiteroses!

whiteroses:
Thanks @bertjwregeer! I've been working on https://github.com/Pylons/pyramid/issues/1264, but the documentation changes are coming!

src/webob/acceptparse.py
```diff
@@ -1859,131 +5102,3 @@ def fdel(request):
             pass
 
     return property(fget, fset, fdel, textwrap.dedent(doc))
-
-
-class MIMEAccept(Accept):
```
bertjwregeer: We should keep this around, even if it is just a shim that calls the new classes/methods. `MIMEAccept` is pretty heavily used by various other libraries/tools that depend on WebOb.

bertjwregeer:
```
MIMEAccept = Accept
```

May be acceptable if the API hasn't changed.

whiteroses:
I think it would need to be

    MIMEAccept = AcceptValidHeader

but the API hasn't changed (other than the new `acceptable_offers` method) &mdash; is that ok? The other libraries/tools don't use `MIMENilAccept`, `NilAccept` or `NoAccept` directly, do they? (These classes were not documented.)

bertjwregeer:
Since they were not documented, I'm less worried. `MIMEAccept` on the other hand was heavily used.

Your suggestion of `MIMEAccept = AcceptValidHeader` would be fine by me.

whiteroses:
Ok, I've made the change.

(Have been wanting to work on the documentation changes and other follow-up work from GSoC, but have been really busy &mdash; may have more time in the second half of this month and in December though.)

ryanpetrello:
I know I'm a bit late to the party on this one, and I applaud you all for the work you've done improving `Accept` header support here.

I work in open source, and I don't mean to be critical or diminish the work that you all are doing (many of you as volunteers, which I also understand).  I know that it can be difficult - and sometimes impossible - to improve software without _some_ breaking API changes.  I _do_ wish, however, that this sort of backwards-incompatible change could have gone into a major version release instead of 1.7 -> 1.8 (though admittedly this could be a matter of semantics, depending what the webob team considers "major versions"), or at the very least involved some form of deprecation period (maybe it did, and I'm not paying enough attention in the right places?); hunting down a bug I was encountering related to this change was fairly painful and surprising, and for now, I've ended up pinning to an earlier version of webob 😢

stevepiercy:
@ryanpetrello You can [watch GitHub repos](https://help.github.com/articles/watching-and-unwatching-repositories/) that affect your software so that you can receive notifications of updates. If that is too noisy, we also make announcements via [Google Group pylons-discuss](https://groups.google.com/d/msg/pylons-discuss/BVOZyw6tzNI/hnYIkYxoBAAJ), and [Twitter](https://twitter.com/PylonsProject/status/947847674203648000). And there is a well-maintained changelog on [PyPI](https://pypi.python.org/pypi/WebOb/1.8.0rc1) and WebOb's [docs](https://docs.pylonsproject.org/projects/webob/en/master/#change-history).

If you have suggestions for how to mitigate your pain and surprise apart from the foregoing options, please let us know.

bertjwregeer:
@ryanpetrello I put out a call for testing for 1.8, including finding out what was broken. I can't test all software that relies on WebOb. What in these changes caused you pain? What bug did it cause? If I knew what it was that broke, I may be able to find a way to provide a shim or the necessary compatibility to allow your code to continue to function.

Also, WebOb like Pyramid and many other Pylons Project projects do not use semantic versioning. And 1.7 -> 1.8 may deprecate public API's (which should at least warn about the upcoming deprecation). This part of the code base was some of the most hairy and difficult to make a transition without potential breaking changes. AFAIK the public/documented API did not change (other than having a deprecation warning on them), internals are allowed to be changed (and have) as needed.

whiteroses:
@ryanpetrello We had many discussions at the time about what to change and where to keep backward compatibility (e.g. under [this PR](https://github.com/Pylons/webob/pull/335)), and tried to keep any changes to the documented public API to a minimum (it's been a while, but I can't actually remember any significant changes to the documented API). You can see we've put in quite a few deprecation warnings for future changes that haven't happened yet. If there's an oversight or mistake on my part, please let us know and I'll (with @bertjwregeer's agreement) do my best to fix.

ryanpetrello:
Hey @whiteroses @bertjwregeer - thanks for the replies.  It looks like you all did a great job broadcasting these coming changes to the community, and that I just wasn't paying attention in the right channels.  I've since subscribed to some of the resources you've mentioned and will pay closer attention in the future.

One specific aspect of this change seems to be tripping up code like this:

`getattr(req.accept, 'header_value', '*/*')`

It looks like this no longer behaves the same way for `AcceptInvalidHeader` and `AcceptNoHeader` as it did before with `webob.acceptparse.MIMEAccept`; now it evaluates to `None`, and that's what was causing issues for me.

> internals are allowed to be changed (and have) as needed, if you rely on reading the WebOb source code and grabbing internal values we can't continue to support that.

I think, in the end, this is what it comes down to - you all made a change to the underlying implementation, and I had code that was relying on it.  I think I've worked out a way to support what I'm doing in both webob 1.8 _and_ prior, so I'm good to go on my end; thanks @bertjwregeer for offering to help me work it out.

bertjwregeer:
Excellent. I was actually just now looking through Pecan since I saw the pin commit come by, I assume you've also found the docs for the new classes in 1.8 here:

https://docs.pylonsproject.org/projects/webob/en/1.8-branch/api/webob.html#webob.acceptparse.AcceptValidHeader.acceptable_offers

In the case of no header being provided, the order that the server wishes to respond in will be taken:

https://docs.pylonsproject.org/projects/webob/en/1.8-branch/api/webob.html#webob.acceptparse.AcceptNoHeader.acceptable_offers

This allows the server to very easily even in the case of no accept header or `*/*` to select the appropriate offer using:

```
clientaccepts = req.accept.acceptable_offers(['text/html', 'application/json'])
```

To get back a list of tuples.

Unless there is anything major, I will be releasing 1.8 within the next day or two. Since you mentioned you have a work-around that works for both 1.8 and older, I won't let this hold up the publishing of the new version.

PR #344, opened 2018-01-17
add trailing whitespaces to deprecation warning for MIMEAccept

stevepiercy:


bertjwregeer:
Could you please put the docs changes into a separate PR?

bertjwregeer:
Create a new branch off master, cherry-pick the add trailing whitespace commit onto it, and create a new PR with that.

Then do the same for the docs change.

So that there are two separate branches, and two separate PR's.

stevepiercy:
After much flailing, including deleting my master branch along with those commits, and manually redoing the changes, I think I got it corrected. What fun! I couldn't figure it out otherwise with my crappy git skilz. See #349 and #350. Did I do this right?

As an aside, are you now enforcing the practice of creating new branches for all PRs for webob? Maybe I've been privileged by assuming I could just make PRs from master, like I do for all other repos under Pylons, and I should adapt. We've been kind of loose about that.

bertjwregeer:
I personally much prefer a cleaner history where each commit can be traced to a PR or branch. I avoid doing code changes directly on master.

I am not against it, but I feel like good commit hygiene makes things simpler in the future if we need to track what happened and why!

Those two PR's look good!

Issue #351, opened 2018-03-06
DELETE-request with 'application/x-www-form-urlencoded' content type.


Cykooz:
Property POST of DELETE-request returns `NoVars` despite the fact that Content-Type is 'application/x-www-form-urlencoded'.

bertjwregeer:
That's documented as such:

https://github.com/Pylons/webob/blob/master/src/webob/request.py#L755

I'd be okay with a pull request that changes it so that it is supported on any arbitrary method.

bertjwregeer:
Closed by #352

PR #352, opened 2018-03-08
request.POST now supports any requests with the appropriate Content-Type (#351)

Cykooz:
Issue #351

### DIFF
--- a/src/webob/request.py
+++ b/src/webob/request.py
@@ -752,12 +752,10 @@ class BaseRequest(object):
         Return a MultiDict containing all the variables from a form
         request. Returns an empty dict-like object for non-form requests.
 
-        Form requests are typically POST requests, however PUT, PATCH & DELETE
+        Form requests are typically POST requests, however any other
         requests with an appropriate Content-Type are also supported.
         """
         env = self.environ
-        if self.method not in ('POST', 'PUT', 'PATCH', 'DELETE'):
-            return NoVars('Not a form request')
         if 'webob._parsed_post_vars' in env:
             vars, body_file = env['webob._parsed_post_vars']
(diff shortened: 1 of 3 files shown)


bertjwregeer:
Instead of limiting it to just certain request methods, it may be better to just look for content-type if it is not a `POST`, and if it is form-encoded then we can continue.

This way the WebDav request methods also just work.

src/webob/request.py
```diff
@@ -752,11 +752,11 @@ def POST(self):
...
         request. Returns an empty dict-like object for non-form requests.
 
-        Form requests are typically POST requests, however PUT & PATCH requests
-        with an appropriate Content-Type are also supported.
+        Form requests are typically POST requests, however PUT, PATCH & DELETE
+        requests with an appropriate Content-Type are also supported.
         """
         env = self.environ
-        if self.method not in ('POST', 'PUT', 'PATCH'):
+        if self.method not in ('POST', 'PUT', 'PATCH', 'DELETE'):
```
bertjwregeer: I would argue for removal of this entirely, so that it just works on all request methods, Webdav has a bunch of other request methods for example and it doesn't make sense if we disallow `.POST` on those.

As long as we keep the line on 766 we should be good to go.

Cykooz: It is a cool idea. I will remove this condition.

Cykooz:
Ping-ping.

What about merging this PR?

bertjwregeer:
I've been busy with work and the flu. I'm also traveling at the moment. I'll get this merged before I push the release for 1.8.

bertjwregeer:
Thank you!

PR #353, opened 2018-03-16
Fix suggested secret length in doc

hongyuan1306:


stevepiercy:
Pyramid counterpart: https://github.com/Pylons/pyramid/pull/3244

PR #358, opened 2018-04-11
Remove ``key`` kwarg, as it was an oversight.

stevepiercy:
Also I noted this in the docstring for `set_cookie()`:

```rst
        ``value``

           The cookie value, which should be a string or ``None``.  If
           ``value`` is ``None``, it's equivalent to calling the
           :meth:`webob.response.Response.unset_cookie` method for this
           cookie key (it effectively deletes the cookie on the client).
```

Should that last line be "cookie *name*"?

### DIFF
--- a/src/webob/response.py
+++ b/src/webob/response.py
@@ -949,8 +949,7 @@ class Response(object):
 
     def set_cookie(self, name=None, value='', max_age=None,
                    path='/', domain=None, secure=False, httponly=False,
-                   comment=None, expires=None, overwrite=False, key=None,
-                   samesite=None):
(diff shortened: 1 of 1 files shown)


stevepiercy:
Bah, ignore my PR. You're too quick, @bertjwregeer!

However, please take a look at the comment to this PR, as I wasn't sure whether to do anything about that.

bertjwregeer:
Yeah, that last one should be cookie name. Missed that.

stevepiercy:
Want me to submit a PR or do you got it?

stevepiercy:
I got it. PR coming now.

PR #364, opened 2018-07-13
Add Request.remote_host property

nirs:
This property expose the REMOTE_HOST environment variable.


nirs:
Ping?

bertjwregeer:
I have received this. I just haven't had time to merge it.

That being said `remote_host` in almost all cases will likely be the same as `remote_addr`, nor should you ever trust the result in `remote_host` since reverse DNS does not have to be accurate.

PR #371, opened 2018-09-03
do not raise on invalid offers

### LINKED ISSUE #365: validate media types consistently in acceptable_offers
```python
>>> create_accept_header('').acceptable_offers(['foo'])
  File "/Users/michael/work/oss/pyramid/env/lib/python3.6/site-packages/webob/acceptparse.py", line 804, in acceptable_offers
    raise ValueError(repr(offer) + ' is not a media type.')
ValueError: 'foo' is not a media type.
>>> create_accept_header('*/*').acceptable_offers(['foo'])
  File "/Users/michael/work/oss/pyramid/env/lib/python3.6/site-packages/webob/acceptparse.py", line 804, in acceptable_offers
    raise ValueError(repr(offer) + ' is not a media type.')
ValueError: 'foo' is not a media type.
>>> create_accept_header('invalid/x/y').acceptable_offers(['foo'])
[('foo', 1.0)]
>>> create_accept_header(None).acceptable_offers(['foo'])
[('foo', 1.0)]
```

It's not ok to raise here in some scenarios and not in others. The offer is server-side and webob should handle it consistently by validating those parameters the same always.

I went backward and found out that webob < 1.5 actually did this consistently but after that it stopped calling `_check_offer` in `MIMEAccept` but kept calling it in `MIMENilAccept`.

mmerickel:
fixes #365

### DIFF
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -796,18 +796,18 @@ class AcceptValidHeader(Accept):
         lowercased_offers = [offer.lower() for offer in offers]
 
         lowercased_offers_parsed = []
-        for offer in lowercased_offers:
+        for offer_index, offer in enumerate(lowercased_offers):
             match = self.media_type_compiled_re.match(offer)
             # The regex here is only used for parsing, and not intended to
             # validate the offer
             if not match:
-                raise ValueError(repr(offer) + ' is not a media type.')
-            lowercased_offers_parsed.append(match.groups())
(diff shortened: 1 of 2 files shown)


mmerickel:
Closing this as I'm not sure it goes far enough to validate the offers. I'd rather see what @whiteroses had in mind in https://github.com/Pylons/webob/issues/365#issuecomment-417069509

PR #372, opened 2018-09-03
do not raise on offers that do not match the media type ABNF

### LINKED ISSUE #365: validate media types consistently in acceptable_offers

mmerickel:
fixes #365

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -38,5 +38,5 @@ Bugfix
   match the required media type grammar when calling ``.acceptable_offers()``.
   Previous versions raised a ``ValueError`` for invalid offers in
   ``AcceptValidHeader`` and returned them as acceptable in the others.
-  See https://github.com/Pylons/webob/pull/371
+  See https://github.com/Pylons/webob/pull/372
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -33,3 +33,10 @@ Bugfix
   MIMEAccept to behave more like the old version. See
   https://github.com/Pylons/webob/pull/356
 
+- ``acceptparse.AcceptValidHeader``, ``acceptparse.AcceptInvalidHeader``, and
+  ``acceptparse.AcceptNoHeader`` will now always ignore offers that do not
+  match the required media type grammar when calling ``.acceptable_offers()``.
+  Previous versions raised a ``ValueError`` for invalid offers in
+  ``AcceptValidHeader`` and returned them as acceptable in the others.
+  See https://github.com/Pylons/webob/pull/371
+
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -419,11 +419,13 @@ class Accept(object):
         lowercased_offers_parsed = []
         for index, offer in enumerate(offers):
             match = self.media_type_compiled_re.match(offer.lower())
-            # The regex here is only used for parsing, and not intended to
-            # validate the offer
-            if not match:
-                continue
-            lowercased_offers_parsed.append([index] + list(match.groups()))
+            # we're willing to try to match any offer that matches the
+            # media type grammar can parse, but we'll throw out anything
+            # that doesn't fit the correct syntax - this is not saying that
+            # the media type is actually a real media type, just that it looks
+            # like one
+            if match:
+                lowercased_offers_parsed.append([index] + list(match.groups()))
         return lowercased_offers_parsed
 
 
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -1253,7 +1253,7 @@ class _AcceptInvalidOrNoHeader(Accept):
         """
         Return the offers that are acceptable according to the header.
 
-        Any offers that do not match the media type ABNF will be ignored.
+        Any offers that do not match the media type grammar will be ignored.
 
         :param offers: ``iterable`` of ``str`` media types (media types can
                        include media type parameters)
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -412,8 +412,8 @@ class Accept(object):
         """
         Throw out any offers that do not match the media type ABNF.
 
-        :return: A list of ABNF-validated offers split into the format:
-                 ``[offer_index, offer_type_subtype, offer_media_params]``.
+        :return: A list of offers split into the format ``[offer_index,
+                 offer_type_subtype, offer_media_type_params]``.
 
         """
         lowercased_offers_parsed = []
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -408,6 +408,24 @@ class Accept(object):
                 )
         return generator(value=value)
 
+    def _parse_and_normalize_offers(self, offers):
+        """
+        Throw out any offers that do not match the media type ABNF.
+
+        :return: A list of ABNF-validated offers split into the format:
+                 ``[offer_index, offer_type_subtype, offer_media_params]``.
+
+        """
+        lowercased_offers_parsed = []
+        for index, offer in enumerate(offers):
+            match = self.media_type_compiled_re.match(offer.lower())
+            # The regex here is only used for parsing, and not intended to
+            # validate the offer
+            if not match:
+                continue
+            lowercased_offers_parsed.append([index] + list(match.groups()))
+        return lowercased_offers_parsed
+
 
 class AcceptValidHeader(Accept):
     """
@@ -771,6 +789,8 @@ class AcceptValidHeader(Accept):
         This uses the matching rules described in :rfc:`RFC 7231, section 5.3.2
         <7231#section-5.3.2>`.
 
+        Any offers that do not match the media type grammar will be ignored.
+
         :param offers: ``iterable`` of ``str`` media types (media types can
                        include media type parameters)
         :return: A list of tuples of the form (media type, qvalue), in
@@ -793,21 +813,12 @@ class AcceptValidHeader(Accept):
             for media_range, qvalue, media_type_params, extension_params in
             parsed
         ]
-        lowercased_offers = [offer.lower() for offer in offers]
-
-        lowercased_offers_parsed = []
-        for offer in lowercased_offers:
-            match = self.media_type_compiled_re.match(offer)
-            # The regex here is only used for parsing, and not intended to
-            # validate the offer
-            if not match:
-                raise ValueError(repr(offer) + ' is not a media type.')
-            lowercased_offers_parsed.append(match.groups())
+        lowercased_offers_parsed = self._parse_and_normalize_offers(offers)
 
         acceptable_offers_n_quality_factors = {}
         for (
-            offer_index, (offer_type_subtype, offer_media_type_params)
-        ) in enumerate(lowercased_offers_parsed):
+            offer_index, offer_type_subtype, offer_media_type_params
+        ) in lowercased_offers_parsed:
             offer_media_type_params = self._parse_media_type_params(
                 media_type_params_segment=offer_media_type_params,
             )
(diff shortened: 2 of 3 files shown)


CHANGES.txt
```diff
@@ -33,3 +33,10 @@ Bugfix
   MIMEAccept to behave more like the old version. See
   https://github.com/Pylons/webob/pull/356
 
+- ``acceptparse.AcceptValidHeader``, ``acceptparse.AcceptInvalidHeader``, and
+  ``acceptparse.AcceptNoHeader`` will now always ignore offers that do not
+  match the required media type ABNF when calling ``.acceptable_offers()``.
```
stevepiercy: First hit:
https://duckduckgo.com/?q=ABNF

abnf.net - Association of Black Nursing Faculty

Suggest a reference for the acronym ABNF:

```
:rfc:`5234`
```

which Sphinx will resolve to https://tools.ietf.org/html/rfc5234

mmerickel: that's fine, fixed in the public api and changelog

CHANGES.txt
```diff
@@ -33,3 +33,10 @@ Bugfix
   MIMEAccept to behave more like the old version. See
   https://github.com/Pylons/webob/pull/356
 
+- ``acceptparse.AcceptValidHeader``, ``acceptparse.AcceptInvalidHeader``, and
+  ``acceptparse.AcceptNoHeader`` will now always ignore offers that do not
+  match the required media type grammar when calling ``.acceptable_offers()``.
+  Previous versions raised a ``ValueError`` for invalid offers in
+  ``AcceptValidHeader`` and returned them as acceptable in the others.
+  See https://github.com/Pylons/webob/pull/371
```
whiteroses: Is this URL intended to be the URL of this PR (372 instead of 371)?

mmerickel: fixed, thanks

src/webob/acceptparse.py
```diff
@@ -408,6 +408,24 @@ def generator(value):
...
+
+        :return: A list of ABNF-validated offers split into the format:
+                 ``[offer_index, offer_type_subtype, offer_media_params]``.
+
+        """
+        lowercased_offers_parsed = []
+        for index, offer in enumerate(offers):
+            match = self.media_type_compiled_re.match(offer.lower())
+            # The regex here is only used for parsing, and not intended to
+            # validate the offer
```
whiteroses: This comment is cryptic and confusing (sorry about that) &mdash; could we replace it with something like `Although the offers do not *need* to be validated, by parsing the offers they are necessarily validated in the process. So we drop the offers that are invalid:`

src/webob/acceptparse.py
```diff
@@ -408,6 +408,24 @@ def generator(value):
                 )
         return generator(value=value)
 
+    def _parse_and_normalize_offers(self, offers):
```
whiteroses: Could I ask what "normalize" means here?

mmerickel: mostly just lowercase

mmerickel: but also guarantee that everything that is returned passes the media type abnf, so the function is doing parse/filter/lowercase.

src/webob/acceptparse.py
```diff
@@ -408,6 +408,24 @@ def generator(value):
...
+
+        """
+        lowercased_offers_parsed = []
+        for index, offer in enumerate(offers):
+            match = self.media_type_compiled_re.match(offer.lower())
+            # The regex here is only used for parsing, and not intended to
+            # validate the offer
+            if not match:
+                continue
+            lowercased_offers_parsed.append([index] + list(match.groups()))
```
whiteroses: I think the above 3 lines can be replaced with

    if match:
        lowercased_offers_parsed.append([index] + list(match.groups()))

src/webob/acceptparse.py
```diff
@@ -408,6 +408,24 @@ def generator(value):
                 )
         return generator(value=value)
 
+    def _parse_and_normalize_offers(self, offers):
+        """
+        Throw out any offers that do not match the media type ABNF.
+
+        :return: A list of ABNF-validated offers split into the format:
+                 ``[offer_index, offer_type_subtype, offer_media_params]``.
```
whiteroses: I think `offer_media_params` should be `offer_media_type_params`?

src/webob/acceptparse.py
mmerickel: fixed

src/webob/acceptparse.py
mmerickel: fixed by other wording

src/webob/acceptparse.py
mmerickel: fixed

src/webob/acceptparse.py
```diff
@@ -771,6 +791,8 @@ def acceptable_offers(self, offers):
         This uses the matching rules described in :rfc:`RFC 7231, section 5.3.2
         <7231#section-5.3.2>`.
 
+        Any offers that do not match the media type grammar will be ignored.
```
bertjwregeer: grammar feels weird to me in this context, but I don't know what would be better here!

src/webob/acceptparse.py
```diff
@@ -1250,7 +1263,11 @@ def acceptable_offers(self, offers):
                  where each offer in `offers` is paired with the qvalue of 1.0,
                  in the same order as in `offers`.
         """
-        return [(offer, 1.0) for offer in offers]
+        return [
+            (offers[offer_index], 1.0)
+            for offer_index, _, _
+            in self._parse_and_normalize_offers(offers)
+        ]
```
whiteroses: Could we add a comment here to explain why we are parsing even though it might seem unnecessary? Something like `We call _parse_and_normalize_offers() here so that invalid offers are dropped as in  AcceptValidHeader.acceptable_offers.`?

src/webob/acceptparse.py
bertjwregeer: Normalize in computing tends to be about making sure that input is valid and is in a consistent form so that further operations can rely on the data being "clean".

See https://en.wikipedia.org/wiki/Text_normalization

src/webob/acceptparse.py
mmerickel: fixed

src/webob/acceptparse.py
mmerickel: well steve didn't like abnf, you guys can duke it out but I'm fine with it the way it is.

tests/test_acceptparse.py
```diff
@@ -909,16 +909,30 @@ def test_accepts_html(self, header_value, returned):
...
+        (
+            AcceptInvalidHeader('foo'),
+            ['text/html', 'text/plain;p=1;q=0.5;e=1', 'foo'],
+            [('text/html', 1.0)],
+        ),
+        (
+            AcceptNoHeader(),
+            ['text/html', 'text/plain;p=1;q=0.5;e=1', 'foo'],
+            [('text/html', 1.0)],
+        ),
```
whiteroses: This test is under the test class `TestAcceptValidHeader` &mdash; could the `AcceptNoHeader` and `AcceptInvalidHeader` tests be moved to their respective test classes? And if we create the header within the test as in the other tests, we can remove the `header` test parameter, save on repetition, and the `AcceptNoHeader` and `AcceptInvalidHeader` can then be tested with the exactly same four cases as well?

mmerickel: I'd be fine with moving the test to somewhere else but I want the examples kept together because the point of this test is to make sure that all 3 variants filter invalid offers in the same way.

src/webob/acceptparse.py
mmerickel: this is merged, do you want to just make a PR to fix these things?

whiteroses: Ah, I'm too slow &mdash; ok, will do!

tests/test_acceptparse.py
whiteroses: But we could say that about all the other tests in the module where the behaviour needs to be consistent across the three header classes, so all the tests would need to be bound together. If you want that to be changed for all the tests to make sure they don't go out-of-sync with each other, that can be done, but as things are right now this test stands out as completely different from how all the tests around it are done and organised, such that we (or anyone else) wouldn't know to look for the tests for `AcceptInvalidHeader` and `AcceptNoHeader` under `TestAcceptValidHeader`.

Or if this is uniquely important, I could submit a PR to move the tests to their own test class?

whiteroses:
Thank you for fixing this @mmerickel!

tests/test_acceptparse.py
mmerickel: it made sense to me to group them, but if it's inconsistent with your test suite then I think it's fine if you want to break them apart

bertjwregeer:
```import pytest

numbers = [1,2,3,4,5]
vowels = ['a','e','i','o','u']
consonants = ['x','y','z']


@pytest.mark.parametrize('number', numbers)
@pytest.mark.parametrize('vowel', vowels)
@pytest.mark.parametrize('consonant', consonants)
def test(number, vowel, consonant):
    pass```

Is a valid way to write a test...

whiteroses: @bertjwregeer But is that still binding the tests for `AcceptValidHeader`, `AcceptNoHeader` and `AcceptInvalidHeader` together in `TestAcceptValidHeader`? I was just going to place the tests in `TestAcceptValidHeader`, `TestAcceptNoHeader` and `TestAcceptInvalidHeader`, as has been done for all the other tests in the rest of the module. I understand the reasoning behind keeping the tests together to ensure a consistent API, but then we'd need to go through all the other tests for all the four headers and do the same?

bertjwregeer: That's fine for now, but we may be able to have some tests that are collapsed so that we can make sure there are no regressions moving forward where someone accidentally changes something in 2 of the 3 for example.

whiteroses: I submitted the PR to move the tests into their separate test classes for now. Should we open an issue for binding the tests together? If you don't consider it urgent, I can work on it when I have time. I'm thinking move all Accept-header-related tests into one TestAccept class, and test 2 or 3 of the classes together where appropriate &mdash; is that what you have in mind?

mmerickel: I think the grouping on these few tests was just to ensure consistency on the same inputs being filtered the same way. I don't have a strong opinion on reorganizing the rest of the suite.

whiteroses: It seems like it'd be a good idea, as we need the api to be consistent across the three classes, and I think there are more methods like this one that would benefit from having the tests be grouped together. There are other, more urgent issues right now that I'd like to work on first, but I can submit a PR at some point so you can see if you think it helps.

PR #376, opened 2018-09-13
support a standard api for parsing media types

mmerickel:


mmerickel:
This is ready for review. I can remove the specificity stuff from `AcceptOffer` if desired... it's not used anywhere but isn't really hurting anything and the conceptual ordering is used in a lot of places. I chose to not add a separate PR for a bug fix with `acceptable_offers` where `Accept: */*` did not match a range-based offer.

src/webob/acceptparse.py
```diff
@@ -407,25 +432,70 @@ def generator(value):
...
+        #    == *        == *           true         Y/N
+        #      N           N              N           N      a/b
+        #      N           N              Y           N      a/b;x=y
+        #      N           Y              N           N      a/*
+        #      N           Y              Y           Y      a/*;x=y
+        #      Y           N              N           Y      */b
+        #      Y           N              Y           Y      */b;x=y
+        #      Y           Y              N           N      */*
+        #      Y           Y              Y           Y      */*;x=y
+        # simplifies to (A and not B or B and C)
```
stevepiercy: Would this table be useful in the [narrative docs](https://docs.pylonsproject.org/projects/webob/en/latest/reference.html#accept-headers)?

I don't understand the "simplifies" comment. Does it pertain to the table's "example" column?

stevepiercy: Do A, B, and C map to the columns, the example, or something else?

mmerickel: yes

mmerickel:
This PR now incorporates #378 and removes media range support from `parse_offer` and subsequently `acceptable_offers`.

PR #379, opened 2018-10-10
support passing around AcceptOffer objects

mmerickel:


src/webob/acceptparse.py
```diff
@@ -867,7 +899,7 @@ def acceptable_offers(self, offers):
                     offer_type == range_type
                     and offer_subtype == range_subtype
                 ):
-                    if range_media_type_params == []:
+                    if range_media_type_params == ():
                         # If offer_media_type_params == [], the offer and the
```
bertjwregeer: Comments needs updating.

bertjwregeer: Meh, whatever.

PR #381, opened 2018-10-15
Start using Black for code formatting

bertjwregeer:


src/webob/multidict.py
```diff
@@ -26,12 +21,13 @@ class MultiDict(MutableMapping):
 
     def __init__(self, *args, **kw):
         if len(args) > 1:
-            raise TypeError("MultiDict can only be called with one positional "
-                            "argument")
+            raise TypeError(
+                "MultiDict can only be called with one positional " "argument"
```
bertjwregeer: Interesting... surprised it doesn't join these...

PR #399, opened 2019-06-02
Support for Accept specificity

kjmph:
In the case that clients send "Accept" header fields that indicate a
preferential response media type, the acceptable_offers function was
dropping the preference.

This change introduces a stable sort on specificity, as well as normal
ordering. This is accompanied by a new test.

### DIFF
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -1008,7 +1008,7 @@ class AcceptValidHeader(Accept):
         acceptable_offers_n_quality_factors.sort(key=lambda tuple_: tuple_[2])
         # (stable) sort by qvalue and then specificity, descending
         acceptable_offers_n_quality_factors.sort(
-            key=lambda tuple_: (tuple_[1], tuple_[3], reverse=True
+            key=lambda tuple_: (tuple_[1], tuple_[3]), reverse=True
         )
         # drop offer_index, specificity
         acceptable_offers_n_quality_factors = [
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -1006,13 +1006,9 @@ class AcceptValidHeader(Accept):
         ]
         # sort by offer_index, ascending
         acceptable_offers_n_quality_factors.sort(key=lambda tuple_: tuple_[2])
-        # (stable) sort by specificity, descending
+        # (stable) sort by qvalue and then specificity, descending
         acceptable_offers_n_quality_factors.sort(
-            key=lambda tuple_: tuple_[3], reverse=True
-        )
-        # (stable) sort by qvalue, descending
-        acceptable_offers_n_quality_factors.sort(
-            key=lambda tuple_: tuple_[1], reverse=True
+            key=lambda tuple_: (tuple_[1], tuple_[3], reverse=True
         )
         # drop offer_index, specificity
         acceptable_offers_n_quality_factors = [
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -993,8 +993,9 @@ class AcceptValidHeader(Accept):
                 )
 
         acceptable_offers_n_quality_factors = [
-            # key is offer, value[0] is qvalue, value[1] is offer_index
-            (key, value[0], value[1])
+            # key is offer, value[0] is qvalue, value[1] is offer_index,
+            # value[2] is specificity
+            (key, value[0], value[1], value[2])
             for key, value in acceptable_offers_n_quality_factors.items()
             if value[0]  # != 0.0
             # We have to filter out the offers with qvalues of 0 here instead
@@ -1005,11 +1006,15 @@ class AcceptValidHeader(Accept):
         ]
         # sort by offer_index, ascending
         acceptable_offers_n_quality_factors.sort(key=lambda tuple_: tuple_[2])
+        # (stable) sort by specificity, descending
+        acceptable_offers_n_quality_factors.sort(
+            key=lambda tuple_: tuple_[3], reverse=True
+        )
         # (stable) sort by qvalue, descending
         acceptable_offers_n_quality_factors.sort(
             key=lambda tuple_: tuple_[1], reverse=True
         )
-        # drop offer_index
+        # drop offer_index, specificity
         acceptable_offers_n_quality_factors = [
             (item[0], item[1]) for item in acceptable_offers_n_quality_factors
         ]
--- a/tests/test_acceptparse.py
+++ b/tests/test_acceptparse.py
@@ -1162,6 +1162,8 @@ class TestAcceptValidHeader(object):
             ),
             ("*/*", ["text/*"], []),
             ("", ["text/*", "*/*", "text/html", "text/html;level=1", "image/*"], []),
+            ("application/json, */*", ["text/html", "application/json"],
+             [("application/json", 1.0), ("text/html", 1.0)]),
         ],
     )
     def test_acceptable_offers__valid_offers(
--- a/tests/test_acceptparse.py
+++ b/tests/test_acceptparse.py
@@ -1162,8 +1162,11 @@ class TestAcceptValidHeader(object):
             ),
             ("*/*", ["text/*"], []),
             ("", ["text/*", "*/*", "text/html", "text/html;level=1", "image/*"], []),
-            ("application/json, */*", ["text/html", "application/json"],
-             [("application/json", 1.0), ("text/html", 1.0)]),
+            (
+                "application/json, */*",
+                ["text/html", "application/json"],
+                [("application/json", 1.0), ("text/html", 1.0)],
+            ),
         ],
     )
     def test_acceptable_offers__valid_offers(
--- a/tests/test_request.py
+++ b/tests/test_request.py
@@ -2621,8 +2621,8 @@ class TestRequest_functional(object):
         ]
         req = self._blankOne("/", accept=["text/plain", "message/*"])
         assert fut(req, ["message/x-foo", "text/plain"]) == [
-            ("message/x-foo", 1.0),
             ("text/plain", 1.0),
+            ("message/x-foo", 1.0),
         ]
         assert fut(req, ["text/plain", "message/x-foo"]) == [
             ("text/plain", 1.0),
--- a/tests/test_request.py
+++ b/tests/test_request.py
@@ -2612,8 +2612,8 @@ class TestRequest_functional(object):
         assert req.accept
         req = self._blankOne("/", accept=["*/*", "text/*"])
         assert fut(req, ["application/x-foo", "text/plain"]) == [
-            ("application/x-foo", 1.0),
             ("text/plain", 1.0),
+            ("application/x-foo", 1.0),
         ]
         assert fut(req, ["text/plain", "application/x-foo"]) == [
             ("text/plain", 1.0),


bertjwregeer:
Hey @whiteroses, would you be so kind as to give your feedback on this? It looks like a very small minimal change for the better, but does this match up with the expectations of the spec?

src/webob/acceptparse.py
```diff
@@ -1005,11 +1006,15 @@ def acceptable_offers(self, offers):
         ]
         # sort by offer_index, ascending
         acceptable_offers_n_quality_factors.sort(key=lambda tuple_: tuple_[2])
+        # (stable) sort by specificity, descending
+        acceptable_offers_n_quality_factors.sort(
```
bertjwregeer: Wouldn't this sort override the sort above?

kjmph: Great question, as it is a stable sort, it won't impact the ordering above unless the specificity has changed. For instance:

```
>>> a = create_accept_header('application/json, text/html')
>>> a.acceptable_offers(a, ['text/html', 'application/json'])
[('text/html', 1.0), ('application/json', 1.0)]
```

The ordering of text/html is still first, as order_index took precedence. However, if the specificity differs then specificity wins:

```
>>> a = create_accept_header('application/json, text/*')
>>> a.acceptable_offers(a, ['text/html', 'application/json'])
[('application/json', 1.0), ('text/html', 1.0)]
```
For completeness, in the event that quality scores outrank specificity, then quality is the final tie breaker:

```
>>> a = create_accept_header('application/json;q=0.1, text/*;q=1')
>>> a.acceptable_offers(a, ['text/html', 'application/json'])
[('text/html', 1.0), ('application/json', 0.1)]
```

Does this help?

mmerickel: I'm actually surprised to see we are re-sorting the list several times instead of defining a single sort tuple. I must have missed that in the original review. @whiteroses can you help me understand this?

kjmph: The first sort is ascending, the second sort is descending. Which is why it is split into two. However, I did introduce the third sort in this pull request. You are correct that the code change could be one tuple sort:

```
    # (stable) sort by qvalue and then specificity, descending
    acceptable_offers_n_quality_factors.sort(
        key=lambda tuple_: (tuple_[1], tuple_[3]), reverse=True
    )
```

I'll make the change.

bertjwregeer:
@kjmph it looks like you pushed some bad syntax.

kjmph:
Sorry @bertjwregeer, development env a bit awkward, as code change was on a different repo. I didn't use patch for such a simple change. And.. typo. It is fixed now.

src/webob/acceptparse.py
mmerickel: Sorting using a tuple is fundamentally different from doing the separate sorts. My original question for @whiteroses was why it is being done as independent sorts.

mmerickel: Also to be clear you can do ascending and descending sorts with the same tuple by negating some of the entries in the tuple. The decision to use a tuple versus independent sorts should not be dependent on that perceived limitation.

Almost certainly the independent sorts is not correct and it should’ve been done using a tuple so that the subsequent sorts could not move items across regions in the result.

whiteroses:
From my current understanding (in the middle of preparing for critical exams and have very little time at the moment until the 14th, apologies): I think the dropping of the specificity before the ordering was intentional. The section in the RFC on this

> Media ranges can be overridden by more specific media ranges or specific media types.  If more than one media range applies to a given type, the most specific reference has precedence...

...I think only applies when choosing the media range to apply to an offer; so

    >>> create_accept_header("text/*, text/html").acceptable_offers(["text/html", "application/json"])
    [('text/html', 1.0)]

But if I understand correctly, this PR is about the behaviour here:

    >>> create_accept_header("application/json, text/*").acceptable_offers(["text/html", "application/json"])
    [('text/html', 1.0), ('application/json', 1.0)]

From my reading of the RFC, the header is saying, "I will accept application/json or any media type starting with text/, _with equal preference for each media type_" &mdash; in which case it should be left to the application to decide the best media type with which to respond. At least that is what I am able to gather with the time I have at the moment &mdash; if it is not urgent, I can give it much more attention and give you better feedback on the 14th.

src/webob/acceptparse.py
whiteroses: The two sorts can be combined into one by using a tuple and negating the value that needs to be sorted in descending order, which may be more efficient and/or elegant. I don't think the current behaviour is incorrect though, given that .sort() is a stable sort &mdash; could you help me understand why it is incorrect @mmerickel?

kjmph:
I concur with you @whiteroses, Python guarantees stability in sorts. I don't understand the problem with the two sorts versus a single sort. As long as the ordering is preserved. For instance:

```
x.sort(key=lambda t: t[1])
x.sort(key=lambda t: t[0])
```

Is identical to:

```
x.sort(key=lambda t: (t[0], t[1]))
```

As well, perhaps it would be helpful to review the Apache documents?

https://httpd.apache.org/docs/trunk/content-negotiation.html#better

The proposed pull request would match Apache ordering, if we can agree that the client preference is more important than the server preference, as Apache believes.

Best of luck in your exams, this can wait on my end, if this is a distraction. I can always use the public fork until we merge.

mmerickel:
> From my reading of the RFC, the header is saying, "I will accept application/json or any media type starting with text/, with equal preference for each media type" — in which case it should be left to the application to decide the best media type with which to respond. [Edit: If the client wishes to specify that they prefer application/json, they can use qvalues.]

This decision was intentional and we made changes in Pyramid with the expectation that the server-supplied ordering would resolve ambiguities - not the client-supplied ordering.

src/webob/acceptparse.py
mmerickel: A stable sort is about the input order versus the output order with respect to a specific sort key. It does not provide any such guarantee where items will end up with respect to one another when you change the sort key other than that they will not change position if the new sort key says they're fine. If the new sort key says they should change position then they will, without regard for the previous sort key. Maybe since you are doing the sorts from least-important to most-important it works out but I still don't think it is guaranteed.

I think the big thing for me whether the current code is correct or not is that a single sort would be 1) more efficient and 2) more obvious the intentions.

bertjwregeer:
> As well, perhaps it would be helpful to review the Apache documents?

@kjmph this entire project to fix Accept handling was a Google Summer of Code project that @whiteroses completed, if you would like to go back and read the long discussion threads, you are more than welcome to, and in fact I encourage it to give you an understanding on how long and how much time was spent overhauling this. It is also why I am very careful about accepting new changes that may introduce incompatibilities with previously agreed upon direction/RFCs.

This is the reason also why I asked @whiteroses to respond (and we'll wait till he has completed his school work) as he has the most knowledge about this, having read and implemented what the RFC's state, not just the "Apache documents" (which don't match the RFCs exactly).

As far as I remember, @mmerickel is right in that where the RFC's were ambiguous about what needed to be done, we used the server supplied ordering, specifically because this is what web frameworks, like Pyramid, expect. This means your current change would break expectations for a large consumer of WebOb.

bertjwregeer:
@whiteroses take your time, good luck on exams!

kjmph:
@bertjwregeer Thanks for the summary. Looking forward to a conclusion at a later time.

When requests come in as `application/json, */*` there are cases when the client will attempt to JSON parse the result without checking `Content-Type` in the response headers. HTTP Exceptions will thus cause a cascading exception on the client side when the server exception response is HTML; yes, bad clients are bad. I suppose it narrows down to this one sentence in the RFC: ```Media ranges can be overridden by more specific media ranges or specific media types.``` I can relate to your hesitancy to accept this pull request with such a history behind the implementation. I will sign off for now.

Note, an alternative, of course, is to modify all application code to not use Pyramid's acceptable offer order to determine exception typing, and instead craft exception bodies and set the correct `Content-Type` in the exception handling. It is very convenient to use Pyramid's exceptions however. (Or even more alternatively, Pyramid could allow an application to specify an offer order in the HTTPException class.) I will wait on these options till it is official on the RFC interpretation.

whiteroses:
We did review the Apache documents at the time &mdash; will be able to explain better when I have more time later in the month.

(Thanks for well-wishes for exams!)

src/webob/acceptparse.py
whiteroses: Ok, will submit PR in about 10 days' time! (if it's not urgent)

whiteroses:
Here's what I understand from re-reading the relevant parts of the RFC again (it is *very* ambiguous, so multiple interpretations is entirely understandable).

I think we all agree that this is correct:

    >>> create_accept_header("text/*, text/html").acceptable_offers(["text/html", "application/json"])
    [('text/html', 1.0)]

We can see that specificity has *not* dropped, but has been used to decide between `text/*` and `text/html` (https://github.com/Pylons/webob/blob/master/src/webob/acceptparse.py#L983). Where we differ is on what should happen here:

    >>> create_accept_header("application/json, text/*").acceptable_offers(["text/html", "application/json"])
    [('text/html', 1.0), ('application/json', 1.0)]

The current behaviour is to return `text/html` first. This PR takes the view that `application/json` should be returned first, because it is more specific than `text/*`. It hinges on the interpretation of these two sentences in the [RFC](https://tools.ietf.org/html/rfc7231#section-5.3.2):

> Media ranges can be overridden by more specific media ranges or specific media types.  If more than one media range applies to a given type, the most specific reference has precedence.

I believe they were intended to have the following meaning:

[**When determining the media type quality factor for a media type,**] media ranges can be overridden by more specific media ranges or specific media types[<b>:</b>] if more than one media range applies to a given type, the most specific reference has precedence.

Reasons for this interpretation:

<ul><li>Determination of the best offers is based on the "media type quality factor":

<blockquote>The media type quality factor associated with a given type is determined by finding the media range with the highest precedence that matches the type.</blockquote>

It is *within* this determination of the media type quality factor for each type (each offer) that "Media ranges can be overridden by more specific media ranges or specific media types"; we can see an example of this in the determination of the media type quality factor/quality value for `text/html;level=2` in the table. Hence the next sentence: "if more than one media range applies to a given type, the most specific reference has precedence." This is WebOb's current behaviour.

This PR uses the specificity in the same way --- but then *uses it again* after all the media type quality factors for each offer have been determined, as a tiebreaker. However, there is no mention of anything like this in the RFC after the table: media ranges are only involved in the determination of the media type quality factors, not anything after. This is why I think the full stop after "Media ranges can be overridden by more specific media ranges or specific media types" was intended to be a colon &mdash; then it all makes sense.

This is also in line with the examples in the RFC, where specificity is only ever shown to be used to determine precedence for media types and ranges of the same type (e.g. "text/"), along with `*/*` &mdash; because those are the media types and ranges that an offer might match when determining an offer's media type quality factor.</li>

<li>If the client wishes to specify that they prefer <code>application/json</code> or <em>any</em> single media type under <code>text/*</code> equally, with the current behaviour, they can. And if the client wishes to specify that they prefer <code>application/json</code> over any media type under <code>text/*</code>, they can do so with qvalues. But with the interpretation behind this PR, how would the client be able to specify a header for the first case?</li></ul>

So I think the current behaviour is correct. (I don't actually see anything that disagrees with this in the section of the Apache document you linked to @kjmph &mdash; `text/html` and `text/plain` have precedence over `*/*`, and `image/gif` and `image/jpeg` have precedence over `*/*`, but nothing there says that e.g. `text/html` should have precedence over `image/*`?)

whiteroses:
@kjmph: The problem in the Apache document, if I'm understanding correctly, is this:

"The intention of this is to indicate that the explicitly listed types are preferred, but if a different representation is available, that is ok too."

    >>> create_accept_header("text/html, text/plain, image/gif, image/jpeg, */*").acceptable_offers(["different/representation", "image/gif"])
    [('different/representation', 1.0), ('image/gif', 1.0)]

I think the problem is with the client here. This meaning of wildcards is not in the RFC &mdash; I think it's a confusion with the meaning of wildcards in the other `Accept-` headers, although there may be some history that I am not aware of. It seems to me that if this were common behaviour that were important to support, the update to RFC 7231 from [RFC 2616](https://tools.ietf.org/html/rfc2616#section-14.1) would have included it. It is also easily corrected on the client's end, by specifying a lower qvalue for the wildcard ranges &mdash; whereas under that interpretation, there is no way to specify a header expressing e.g. equal preference for `application/json` or any single media type under `text/*` (as mentioned in the previous comment). So I still think that WebOb's current behaviour is correct.

src/webob/acceptparse.py
whiteroses: @mmerickel Submitted [PR](https://github.com/Pylons/webob/pull/401) to combine the sorts as you suggested. It is much more readable &mdash; not sure what I was thinking. Thanks!

kjmph:
Thank you for the thorough analysis. I also appreciate you following through with the analysis of the Apache document, as well. I agree with you completely that the RFC is ambiguous, and open to interpretation.

When I read through your edited version with the surrounding descriptions, I concur that WebOb is implemented correctly. Based on this, I also agree that this pull requested, if accepted, would break the the ability for the client to specify that they prefer `application/json` or *any* single media type under `text/*` equally. Where as, with the current implementation the qvalues holding equal do express this.

I also conclude (based on this discussion) that clients incorrectly interpret the purpose of this header, with Edge being egregious with its `text/html, application/xhtml+xml, image/jxr, */*` accept header. As I can't fix clients, perhaps when using Pyramid for REST style JSON rendering of responses, I should open a pull request there which allows one to configure accept ordering in HTTP Exceptions?

Again, thank you for your time, this was very helpful. I will close the pull request with this comment.

PR #401, opened 2019-06-16
Combine sorts, plus small improvement.

whiteroses:
Combine sorts as suggested by @mmerickel [here](https://github.com/Pylons/webob/pull/399#discussion_r289913511), plus small improvement by removing an unnecessary assignment.

Thanks!

### DIFF
--- a/src/webob/acceptparse.py
+++ b/src/webob/acceptparse.py
@@ -1004,11 +1004,13 @@ class AcceptValidHeader(Accept):
             # and offers is ['text/html']
         ]
         acceptable_offers_n_quality_factors.sort(
-            key=lambda tuple_: (tuple_[1], -tuple_[2]), reverse=True,
+            key=lambda tuple_: (tuple_[1], -tuple_[2]),
+            reverse=True,
             # descending sort by (qvalue, -offer_index)
         )
         return [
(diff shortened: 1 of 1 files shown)


src/webob/acceptparse.py
```diff
@@ -1003,17 +1003,16 @@ def acceptable_offers(self, offers):
...
+            reverse=True,
+            # descending sort by (qvalue, -offer_index)
         )
-        # drop offer_index
-        acceptable_offers_n_quality_factors = [
-            (item[0], item[1]) for item in acceptable_offers_n_quality_factors
+        return [
+            (item[0], item[1])
+            for item in acceptable_offers_n_quality_factors
+            # dropping offer_index
```
bertjwregeer: Don't put this inside the list comprehension, it currently causes a bug in black to be exposed which may cause issues when formatting the code base.

Just place it on the line before the `return` please.

whiteroses: Do you mean the comment @bertjwregeer?

bertjwregeer: Yes, I mean the comment. See https://github.com/python/black/issues/563 for the issue.

whiteroses: Ok, I've made the change.

bertjwregeer:
Thanks @whiteroses!

PR #404, opened 2019-08-17
Fix linkcheck

stevepiercy:
This PR fixes broken or removes dead URLs in source code and docs. There is one weird one that I finally tracked down from Pyramid when building its docs in c6b6237.

### DIFF
--- a/src/webob/request.py
+++ b/src/webob/request.py
@@ -313,7 +313,7 @@ class BaseRequest(object):
         "int",
     )
     remote_user = environ_getter("REMOTE_USER", None)
-    remote_host = environ_getter("REMOTE_HOST", None, "4.1.9")
+    remote_host = environ_getter("REMOTE_HOST", None)
     remote_addr = environ_getter("REMOTE_ADDR", None)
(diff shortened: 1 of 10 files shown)


src/webob/request.py
```diff
@@ -313,7 +313,7 @@ def body_file_seekable(self):
         "int",
     )
     remote_user = environ_getter("REMOTE_USER", None)
-    remote_host = environ_getter("REMOTE_HOST", None, "4.1.9")
+    remote_host = environ_getter("REMOTE_HOST", None)
```
bertjwregeer: This is already fixed in a different branch. Please remove this change for right now.

stevepiercy: Done in d0eda05.

bertjwregeer:
My bad on the Google Cloud build... I added that because I wanted to test it, but apparently it just fails the job if you don't have a `cloudbuild.yaml`

stevepiercy:
Appveyor builds fail, too, but only on py2.7. I have no clue.

PR #405, opened 2019-09-16
Fixing decode error 

epifanio:
Fixing issue with UTF-8 strings when 
```
decoding = self.charset or self.default_body_encoding
```
passes the wrong decoding option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ WebOb.egg-info/
 pytest*.xml
 coverage*.xml
 .pytest_cache/
+.idea/
+
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,4 @@ WebOb.egg-info/
 pytest*.xml
 coverage*.xml
 .pytest_cache/
-.idea/
 
--- a/src/webob/response.py
+++ b/src/webob/response.py
@@ -623,7 +623,7 @@ class Response(object):
         body = self.body
         try:
             return body.decode(decoding, self.unicode_errors)
-        except:
+        except UnicodeDecodeError:
             return body.decode("UTF-8", self.unicode_errors)
 
     def _text__set(self, value):
--- a/src/webob/response.py
+++ b/src/webob/response.py
@@ -621,7 +621,10 @@ class Response(object):
             )
         decoding = self.charset or self.default_body_encoding
         body = self.body
-        return body.decode(decoding, self.unicode_errors)
+        try:
+            return body.decode(decoding, self.unicode_errors)
+        except:
+            return body.decode("UTF-8", self.unicode_errors)
 
     def _text__set(self, value):
         if not self.charset and not self.default_body_encoding:
--- a/src/webob/response.py
+++ b/src/webob/response.py
@@ -623,7 +623,7 @@ class Response(object):
         body = self.body
         try:
             return body.decode(decoding, self.unicode_errors)
-        except UnicodeEncodeError:
+        except UnicodeDecodeError:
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -18,3 +18,5 @@ WebOb.egg-info/
 pytest*.xml
 coverage*.xml
 .pytest_cache/
+.idea/
```
stevepiercy: This should go in your user .gitignore, not here.

src/webob/response.py
```diff
@@ -621,7 +621,10 @@ def _text__get(self):
             )
         decoding = self.charset or self.default_body_encoding
         body = self.body
-        return body.decode(decoding, self.unicode_errors)
+        try:
+            return body.decode(decoding, self.unicode_errors)
+        except UnicodeDecodeError:
+            return body.decode("UTF-8", self.unicode_errors)
```
stevepiercy: This causes [test coverage to fail](https://travis-ci.com/Pylons/webob/jobs/235825175#L513). 100% test coverage is required in PRs.

.gitignore
epifanio: Thanks, I fixed the relative commit.

src/webob/response.py
epifanio: am not familiar with coverage test, trying to learn. I had a look at the code in [test_response.py](https://github.com/Pylons/webob/blob/master/tests/test_response.py#L758-L787)
which seems the part of coverage addressing the function I modified. 
do you have any hints on what needs to be added, in order to fix the coverage issue?
Thanks!

stevepiercy: You need to add a test that [covers that which is not covered](https://travis-ci.com/Pylons/webob/jobs/235825175#L508), specifically the `except` block.

stevepiercy: @epifanio to save time, you can also run tests locally before committing and pushing to this PR by installing `tox` and using `tox -e py27,py36,coverage`. Let me know if you need help with that.

epifanio: Got the `tox -e py27,py36,coverage` working. From the test coverage codei see some clue can come from the following lines  https://github.com/Pylons/webob/blob/master/tests/test_response.py#L758-L787. — it is my understanding I should add a new test method like `def test_text_get_wrong_encoding()`. To try to cover the case when a wrong encoding is passed to `_text_get` method. Is that correct?

stevepiercy: You're on the right path. I would narrow it down by running a specific test in debug mode, and step through, to see which lines are executed. Saves guesswork.

As far as its name, naming is hard. Maybe incorporate the method name, e.g., `text_get_body_unicode_decode_error` or something? Although longer, it gives hints to what it tests.

.gitignore
```diff
@@ -18,3 +18,4 @@ WebOb.egg-info/
 pytest*.xml
 coverage*.xml
 .pytest_cache/
+
```
bertjwregeer: Please don't update the `.gitignore`, I would ask you to please rebase your changes to avoid touching this file at all.

src/webob/response.py
bertjwregeer: We should not fall back to attempting to decode with UTF-8 if the remote client has sent a charset.

bertjwregeer:
I am sorry, but I am going to reject this change. If the remote client has sent a charset, we shouldn't attempt to guess and fall back to decoding UTF-8. If the remote has not set a charset, we are correctly using the `self.default_body_encoding`.

Issue #406, opened 2019-11-28
Accepting "None" (string) value of cookie attribute SameSite.


marc1n:
According to the new Cookie spec https://tools.ietf.org/html/draft-west-cookie-incrementalism-00
 all browsers should accept now "None" (string) value of SameSite attribute in Set-Cookie header.

This issue is a high priority because Chrome has already changed the default value of SameSite (see  https://blog.chromium.org/2019/10/developers-get-ready-for-new.html and https://www.chromium.org/updates/same-site).

jvanasco:
I was just about to file this same bug.

My understanding of the spec and usage is that the string should be `None` (initial cap), not `none` (all lowercase).  Additionally, Chrome is forcing this change and Google websites are using "None" (leading capital N) in their own cookies.

marc1n:
According to spec value of SameSite attribute is case insensitive.

bertjwregeer:
I've merged the change into master. Sometime in the near future I'll do another release.

wiewiorek:
Hi, are there any plans to do release soon? This feature is needed!

bertjwregeer:
There were plans, however Google has a neat page that describes what will happen to older browsers that don't support the new spec:

https://www.chromium.org/updates/same-site/incompatible-clients

And this is not just a matter of users not upgrading, this is current software that is running on current devices that will not work.

This is why SameSite support is marked as experimental (it still is not a valid RFC), and why I am really frustrated with Google forcing a change through at such high speed.

Whose going to be responsible for when a new version of WebOb breaks existing sites because their authors are using Google's suggest value of setting cookies with SameSite to None?

wiewiorek:
Hey! Thank you for your reply. I am sharing your concerns.

But I don't see how making possible to use SameSite="None" (string) within WebOb would cause anyone any trouble, if default would still be not to pass SameSite flag at all. Doing so would give a choice to developers to decide weather they would like to use it or not. Without that we are gonna need some dirty hacks to set this. 

What do you think about that?

jvanasco:
Google already has this rolled out out to 50% of beta users, and is still targeting 100% of normal users next month.  It looks like 20% of my traffic is requiring a patched version of webob now; and that's jumping to 60% next month.  The incompatible browsers are an fraction of this traffic to me.

> Whose going to be responsible for when a new version of WebOb breaks existing sites because their authors are using Google's suggest value of setting cookies with SameSite to None?

[...]

PR #407, opened 2019-11-29
Allows samesite='none' in set_cookie

marc1n:
https://github.com/Pylons/webob/issues/406

### DIFF
--- a/src/webob/cookies.py
+++ b/src/webob/cookies.py
@@ -335,8 +335,9 @@ class Morsel(dict):
             if self.samesite:
                 if not self.secure and self.samesite.lower() == b'none':
                     raise ValueError(
-                        'Incompatible arguments: when the samesite attribute equals "none", '
-                        'then the secure attribute must be True"')
+                        'Incompatible cookie attributes: '
+                        'when the samesite equals "none", then the secure must be True"'
+                    )
                 add(b"SameSite=" + self.samesite)
 
         return native_(b"; ".join(result), "ascii")
--- a/src/webob/cookies.py
+++ b/src/webob/cookies.py
@@ -333,6 +333,10 @@ class Morsel(dict):
                 add(b"HttpOnly")
 
             if self.samesite:
+                if not self.secure and self.samesite.lower() == b'none':
+                    raise ValueError(
+                        'Incompatible arguments: when the samesite attribute equals "none", '
+                        'then the secure attribute must be True"')
                 add(b"SameSite=" + self.samesite)
 
         return native_(b"; ".join(result), "ascii")
--- a/src/webob/cookies.py
+++ b/src/webob/cookies.py
@@ -548,7 +548,7 @@ def make_cookie(
 
     ``samesite``
       The 'SameSite' attribute of the cookie, can be either ``"strict"``,
-      ``"lax"``, ``"none"`` or ``None``.
+      ``"lax"``, ``"none"``, or ``None``.
     """
 
     # We are deleting the cookie, override max_age and expires
--- a/src/webob/cookies.py
+++ b/src/webob/cookies.py
@@ -277,8 +277,8 @@ def serialize_cookie_date(v):
 def serialize_samesite(v):
     v = bytes_(v)
 
-    if v.lower() not in (b"strict", b"lax"):
-        raise ValueError("SameSite must be 'Strict' or 'Lax'")
+    if v.lower() not in (b"strict", b"lax", b"none"):
+        raise ValueError("SameSite must be 'strict', 'lax', or 'none'")
 
     return v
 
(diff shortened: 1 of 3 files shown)


src/webob/cookies.py
```diff
@@ -547,8 +547,8 @@ def make_cookie(
       Set a comment on the cookie. Default: ``None``
 
     ``samesite``
-      The 'SameSite' attribute of the cookie, can be either ``"Strict"``,
-      ``"Lax"``, or ``None``.
+      The 'SameSite' attribute of the cookie, can be either ``"strict"``,
+      ``"lax"``, ``"none"`` or ``None``.
```
stevepiercy: Oxford comma ftw! Please add a comma after the first `none`.

stevepiercy:
[Black](https://github.com/psf/black) is required to format code in WebOb.

marc1n:
I've changed case in doc for better visual distinction between "none" and None.

bertjwregeer:
According to the specs if the value `"None"` is used for the same site attribute then the cookie must be marked as `secure` or it will be ignored by browsers.

Do we want to handle that invariant?

bertjwregeer:
@marc1n would you also kindly remove the merge commit by rebasing this work on top of master?

marc1n:
I tried but I failed. My knowledge of git is too limited. What steps should I do to remove the merge commit?

bertjwregeer:
@marc1n you should be able to run:

`git rebase -i a5712d73498be29b03d4d15dfcee21c2ed45ce1a`

when the editor appears, exit the editor without modifying anything.

Then force push the results to your Github repo:

`git push --force-with-lease`

This will overwrite the copy on Github with your fixed local copy, this PR will then automatically update. Thanks!

marc1n:
@bertjwregeer Done, thanks for help!

marc1n:
> According to the specs if the value `"None"` is used for the same site attribute then the cookie must be marked as `secure` or it will be ignored by browsers.
> 
> Do we want to handle that invariant?

I think "yes", but I am not sure which is better: to raise exception if secure=True is not set, or automatically set it to True.

mmerickel:
> to raise exception if secure=True is not set, or automatically set it to True.

I would raise a ValueError explaining the incompatible arguments.

bertjwregeer:
I would be happy to accept this PR with the suggestion that @mmerickel made. I think that is the best way moving forward to make sure that people get the behavior they are expecting.

jvanasco:
This change in Chrome caused a series of issues for us.

While I support this PR, and am very eager for it to be merged/released ASAP, I want to suggest that a future version does not raise an exception after checking the SameSite arguments or their compatibility with the secure flag - as that could suddenly change *again*.

Perhaps an environment variable could be used to toggle between warning with a deprecation/log  or raising an exception.  The default can absolutely be raising an exception, but there should be an option to easily avoid these checks if the cookie spec changes again or a major browser vendor decides to suddenly force a new behavior or keyword value.

I love that WebOb/Pyramid are helping us "do the right thing", but I am worried about future sudden changes by browser vendors.

bertjwregeer:
There's a reason why I marked this as experimental in WebOb. I am all for security improvements, but the fact remains that Google and Chrome are steamrolling the various communities and forcing them to adapt ever changing standards.

The draft that was linked above expired "November 8, 2019" and the fact that Chrome is steam-rolling ahead with the change without further consent from the internet community is absolutely terrible.

bertjwregeer:
I highly doubt that Google is going to relax the requirement for SameSite=None to not also be Secure, so having that check in the codebase is fine by me.

I will not be adding environment checks to WebOb, I may provide a flag in the future that can be set on a module, but no environment variables. That sort of behaviour has bitten me far too many times in the past (looking at oauthlib...) in other libraries that changed core behaviour depending on environment variables that are not well documented/not well known.

jvanasco:
> I will not be adding environment checks to WebOb, I may provide a flag in the future that can be set on a module, but no environment variables. That sort of behaviour has bitten me far too many times in the past (looking at oauthlib...) in other libraries that changed core behaviour depending on environment variables that are not well documented/not well known.

That's a great point. I prefer toggling a flag in a module too, but I've seen more people advocate for environment variables so went with that (also, probably oauthlib).  I'd be happy to submit a draft PR that uses that approach.

I have a bunch of whitelabel sites that communicate with a centralized domain via javascript + iframes (like facebook and twitter buttons). The Chrome change completely broke this functionality and our overall user experience over a very large chunk of traffic.  This was hotfixed with a patch in the deployment routine, but I really hate this approach.

bertjwregeer:
I'd be open to a draft PR that adds that, although I am not sure it is worth adding the change for something that I think is very unlikely to change in the future. We should be moving towards a more HTTPS web anyway, and the Secure flag on cookies should be the norm.

jvanasco:
I'm just going to use the flag to disable checking the value matches a b

jvanasco:
Understood. I was just going to draft a PR to support disabling the check on the samesite value , in case a new one is added or they start tests again (there were some other values tested at one point).

bertjwregeer:
Oh... I guess that could make sense. Sure, start a PR for that change.

PR #409, opened 2019-12-19
disable validating the SameSite value, if/when the cookie spec changes

jvanasco:
PR #407 expanding the list of allowable SameSite values to included the string `"None"`.

This PR introduces a module flag `SAMESITE_VALIDATION` to bypass this check if/when the cookie specification or major browser implementations change again:

        import webob.cookies
        webob.cookies.SAMESITE_VALIDATION = False

        ck = webob.cookies.make_cookie(cookie_name, value, samesite='future')

### DIFF
--- a/tests/test_cookies.py
+++ b/tests/test_cookies.py
@@ -123,8 +123,8 @@ def test_cookie_samesite_none_not_secure():
         c.serialize()
 
 
-def test_cookie_samesite_future():
-    # first default
+def test_cookie_samesite_future__default():
+    # ensure default behavior when unsupported values are provided
     c = cookies.Cookie()
     with pytest.raises(ValueError) as excinfo:
         c[b"foo"] = b"bar"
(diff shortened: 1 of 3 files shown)


stevepiercy:
Please use [Black](https://github.com/psf/black) to format code to fix the [lint failure](https://travis-ci.com/Pylons/webob/jobs/269056259).

jvanasco:
ack! sorry. I didn't realize tests were in a different directory. I only ran it on src before. :(

tests/test_cookies.py
```diff
@@ -123,6 +123,37 @@ def test_cookie_samesite_none_not_secure():
...
+def test_cookie_samesite_future():
+    # first default
+    c = cookies.Cookie()
+    with pytest.raises(ValueError) as excinfo:
+        c[b"foo"] = b"bar"
+        c[b"foo"].samesite = b"Future"
+        c.serialize()
+    assert excinfo.value.args[0] == "SameSite must be 'strict', 'lax', or 'none'"
+
+    try:
```
bertjwregeer: We use pytest, please use the `monkeypatch` fixture to monkeypatch during this test, and use `pytest.raises()` within the test.

Write two tests, one that tests when the flag is set, and one that tests when the flag is not set, feel free to use a pytest parametrize to test both invariants with a single test.

PR #410, opened 2020-03-22
Remove deprecated isAlive method

carlwgeorge:
`threading.Thread.isAlive` is deprecated in Python 3.8, and removed entirely in Python 3.9.  The alternative `is_alive` is available since Python 2.6.

https://bugs.python.org/issue37804

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -62,7 +62,7 @@ def serve():
             log.debug("shutting server down")
             server.shutdown()
             worker.join(1)
-            if worker.isAlive():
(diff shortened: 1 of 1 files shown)


bertjwregeer:
Already done in https://github.com/Pylons/webob/pull/390/commits/95c6141d47e1fa73891a787d70b4f8b70893efce

This is part of https://github.com/Pylons/webob/pull/390 which is removing Python 2 support, and will be the first version of WebOb that supports Python 3.9.

Thanks.

PR #411, opened 2020-05-26
Fix tests on py3.9, and enabling testing for it

mgorny:


bertjwregeer:
This is part of the larger PR here: https://github.com/Pylons/webob/pull/390

Thanks for your submission!

PR #412, opened 2020-06-02
Rewind body_file_raw after parsing in POST

lazorchakp:
Currently, request.POST reads from `wsgi.input` without rewinding the stream on a `webob._parsed_post_vars` cache miss. This means that WSGI middlewares using WebOb could incidentally prevent a downstream WSGI application from obtaining the raw request body if it uses a different framework.

Here, we seek the raw body file back to its start after cgi parses it.

I've included a test that captures the desired behavior, which fails without the new change. If this is an unreasonable use case or intentional behavior, I'd love to learn more about it. Thanks in advance.

### DIFF
--- a/src/webob/request.py
+++ b/src/webob/request.py
@@ -810,6 +810,7 @@ class BaseRequest(object):
                 encoding="utf8",
             )
 
+        self.body_file_raw.seek(0)
         vars = MultiDict.from_fieldstorage(fs)
         env["webob._parsed_post_vars"] = (vars, self.body_file_raw)
         return vars
--- a/tests/test_request.py
+++ b/tests/test_request.py
@@ -613,6 +613,21 @@ class TestRequestCommon(object):
         assert bar.filename == "bar.txt"
         assert bar.file.read() == b'these are the contents of the file "bar.txt"\n'
 
+    @pytest.mark.parametrize("method", ["POST", "PUT", "PATCH", "DELETE"])
+    def test_POST_preserves_body_file(self, method):
+        data = b"var1=value1&var2=value2&rep=1&rep=2"
+        wsgi_input = BytesIO(data)
+        environ = {
+            "wsgi.input": wsgi_input,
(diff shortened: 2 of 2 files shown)


mmerickel:
Middleware that reads the request body needs to put things back where it found them. The WSGI spec is somewhat clear on this point that middleware should do its best to not require any special support from the application.

Similarly, in webob if you're reading from `body_file_raw` then you're exhausting the stateful iterator - a body is never buffered by default and is always opt-in to that behavior (WSGI allows the body to be a raw socket which is not even seekable).

To get around these issues things like pyramid_retry (which supports retrying requests) invokes `request.make_body_seekable()` which explicitly triggers a full read / buffer of the body which can then be rolled back and the environ is updated to replace wsgi.input with this new buffer.

lazorchakp:
Thanks for the quick and detailed response -

I understand what you mean about the middleware's responsibility to preserve state, especially when it comes to wsgi.input. In the case of parsing form data in the request's body, can we put this responsibility on webob? It would be more consistent with the rest of the request wrapper; for example, `request.body` does do the work to maintain wsgi.input's future readability - I see POST as a similar case.

request.POST already calls `make_body_seekable`, so it doesn't seem like there's any harm or danger in resetting the stream at the end of the getter. However, if we decide not to go through with this change, I'm more than happy to rewind the body myself, outside of webob.

mmerickel:
I looked a little further and noticed that `request.body` (which is also the foundation for `request.text` and `request.json_body`) actually does reset the position before returning the body. With this in mind it seems to stand to reason that `request.POST` should do the same. I think I'm actually +1 on this change.

lazorchakp:
sounds great. If there's anything else you need from me to move forward here, just let me know. Thanks again

mmerickel:
Gonna wait to see if @bertjwregeer agrees with my reasoning!

bertjwregeer:
Agreed! Thanks!

PR #423, opened 2020-11-04
use errors=replace on unicode decode/encoding in NestedMultiDict

platinummonkey:
In the event a client does not properly set the charset header, the encoding/decoding steps assume utf8, but will barf on POST bodies that contain non utf8 characters with a UnicodeDecodeError. In these cases, replace these characters and handle in app.

### DIFF
--- a/src/webob/multidict.py
+++ b/src/webob/multidict.py
@@ -80,12 +80,12 @@ class MultiDict(MutableMapping):
                 else:
 
                     def decode(b):
-                        return b.encode("utf8").decode(charset)
+                        return b.encode("utf8", errors='replace').decode(charset, errors='replace')
 
             else:
 
                 def decode(b):
-                    return b.decode(charset)
+                    return b.decode(charset, errors='replace')
 
             if field.filename:
                 field.filename = decode(field.filename)
--- a/src/webob/multidict.py
+++ b/src/webob/multidict.py
@@ -80,12 +80,14 @@ class MultiDict(MutableMapping):
                 else:
 
                     def decode(b):
-                        return b.encode("utf8", errors='replace').decode(charset, errors='replace')
+                        return b.encode("utf8", errors="replace").decode(
+                            charset, errors="replace"
+                        )
 
             else:
 
                 def decode(b):
-                    return b.decode(charset, errors='replace')
+                    return b.decode(charset, errors="replace")
 
             if field.filename:
                 field.filename = decode(field.filename)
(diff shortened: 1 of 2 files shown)


mmerickel:
Can you provide a reproducible example here - ideally a test that fails without this PR?

mmerickel:
I tried out this test case against the unmodified multidict and it passed on python 3.7 and 2.7. I'm not sure it's demonstrating the issue that you're trying to solve.

mmerickel:
From what I understand at this point, I'll say that I don't think we'll want to change the default behavior of webob here because it passes data to an application that wasn't being passed before. It feels like something that belongs in the `request.decode()` domain where a user can specify alternate approaches to decoding the query string and body. The only part that's confusing me a little bit here is that it's focused specifically on the multidict with a multipart body in which each item in the multipart may specify its own content-type.

platinummonkey:
FWIW, we've forked and implemented this fix. This is happening due to clients setting an incorrect charset that what is actually being sent. We noticeably were only seeing this in multidict/multipart body style requests.

Here is a graph showing the rollout for a very specific API endpoint experiencing this exact error:

Error counts for UnicodeDecodeError

bertjwregeer:
The client is sending garbage... and with this change we are silently ignoring that? I would recommend catching the `UnicodeDecodeError` instead and presenting the user with a proper error message that they sent a `400 Bad Request`.

Silently replacing broken data just doesn't feel like a good idea, and is not something I would be willing to add.

What I would be okay with, is changing the default in the `from_fieldstorage` function to be `latin-1` since the default charset according to the RFC2616 for HTTP is `ISO-8859-1`. This should be a no-op for most clients/requests since they are properly encoding the data. I am a little wary though, because I don't know how much of a breaking change that would be for other users that currently expect the data to be UTF-8 coming out of `MultiDict`

bertjwregeer:
Although, reading this RFC for multi-part data: https://tools.ietf.org/html/rfc7578#section-5.1.2 it disagrees.

>   Some applications of this specification will supply a character
>   encoding to be used for interpretation of the multipart/form-data
>   body.  In particular, HTML 5 [W3C.REC-html5-20141028] uses
>
>   o  the content of a "_charset_" field, if there is one;
>
>   o  the value of an accept-charset attribute of the <form> element, if
>      there is one;
>
>   o  the character encoding of the document containing the form, if it
>      is US-ASCII compatible;
>
>   o  otherwise, UTF-8.

PR #438, opened 2022-10-24
Pass the Request.decode errors argument down to transcode_query

jeremylowery:
Currently, when calling Request.decode("encoding", "ignore"), webob still raises a decode error with invalid characters despite passing an "ignore" argument to errors. This fix uses the given argument.


mmerickel (changes requested):
This seems to make sense. Would love to see a test that passes in the ignore value.

Issue #442, opened 2022-11-18
Misleading documentation of AcceptLanguageValidHeader.lookup


romuald:
URL of the documentation: https://docs.pylonsproject.org/projects/webob/en/stable/api/webob.html#webob.acceptparse.AcceptLanguageValidHeader.lookup

`language_tags` is described as being an *iterable*, however this is not strictly true, as passing a dict (with languages as keys) will break the call for example.

See the test bellow for an actual example

I'm guessing the actual type is a *Sequence* instead

```python
from unittest.mock import Mock

from webob.acceptparse import AcceptLanguageValidHeader


def test_lookup_ok():
    languages = ('en', 'fr')

    header = AcceptLanguageValidHeader('fr, *')

    assert header.lookup(languages, default='fallback') == 'fr'


def test_lookup_ko():
    languages = dict.fromkeys(('en', 'fr'))

    for key in languages:
        # Real world: associate with a gettext instance
        languages[key] = Mock()

    header = AcceptLanguageValidHeader('fr, *')

    assert header.lookup(languages, default='fallback') == 'fr'

""" <<< output <<<<<<
____________________ test_lookup_ko ____________________

    def test_lookup_ko():
        known = dict.fromkeys(('en', 'fr'))
    
        for key in known:
            # Real world: associate with a gettext instance
            known[key] = Mock()
    
        header = AcceptLanguageValidHeader('fr, *')
    
>       assert header.lookup(known, default='fallback') == 'fr'

test_accept.py:23: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
.../webob/acceptparse.py:4686: in lookup
    match = best_match(range_=range_.lower())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

range_ = 'fr'

    def best_match(range_):
        subtags = range_.split('-')
        while True:
            for index, tag in enumerate(lowered_tags):
                if tag in not_acceptable_ranges:
                    continue
                    # We think a non-'*' range with q=0 represents only
                    # itself as a tag, and there should be no falling back
                    # with subtag truncation. For example, with
                    # 'en-gb;q=0', it should not mean 'en;q=0': the client
                    # is unlikely to expect that specifying 'en-gb' as not
                    # acceptable would mean that 'en' is also not
                    # acceptable. There is no guidance on this at all in
                    # the RFCs, so it is left to us to decide how it should
                    # work.
    
                if tag == range_:
[...]

PR #460, opened 2024-02-26
consolidate accept classes into AcceptHeader

mmerickel:
- Ensure `AcceptHeader` instances are immutable.
- Merge missing/invalid/valid into the single instance - there were very very minor differences between the classes.
- Add `AcceptHeaderState` for simplified checking of the state.
- Add `enum-tools[sphinx]` dependency for dealing with `AcceptHeaderState`.
- Fix broken dependency on setuptools in docs.


whiteroses:
Hi @mmerickel, just saw your ping - it's been a while, so I'll need to find some time to refamiliarise myself with everything (may take a few days), but will review.

whiteroses:
Just to let you know, I'm almost done with the reviewing - will post comments tomorrow.

CHANGES.txt
```diff
@@ -10,14 +10,21 @@ Feature
...
+  are compatible with ``webob.acceptparse.Accept.acceptable_offers``. This API
+  also normalizes the offer with lowercased type/subtype and parameter names.
   See https://github.com/Pylons/webob/pull/376 and
   https://github.com/Pylons/webob/pull/379
 
+- Consolidation of ``Accept`` header handling into a single class.
+  See backward incompatibilities below for more information.
+  See https://github.com/Pylons/webob/pull/460
+
+- ``webob.acceptparse.Accept``, methods ``best_match``, ``quality``, and
```
whiteroses, line 23: Would "webob.acceptparse.Accept's methods...", without the comma, be clearer?

CHANGES.txt
```diff
@@ -27,6 +34,28 @@ Backwards Incompatibilities
...
+  attribute for users that need to know the exact state of the header.
+  See https://github.com/Pylons/webob/pull/460
+
+- Remove previously-deprecated ``webob.acceptparse.MIMEAccept``.
+  See https://github.com/Pylons/webob/pull/460
+
+- Remove previously-deprecated ``webob.acceptparse.Accept.__iter__``.
+  See https://github.com/Pylons/webob/pull/460
+
+- ``webob.acceptparse.Accept`` methods, ``best_match``, ``quality``, and
```
whiteroses, line 48: Would "webob.acceptparse.Accept's methods best_match, ..." be clearer?

CHANGES.txt
```diff
@@ -27,6 +34,28 @@ Backwards Incompatibilities
...
+  ``__contains__`` are now thin wrappers around ``acceptable_offers`` which
+  modifies their behavior slightly:
+
+  - Offers containing wildcards are no longer allowed.
+  - A tuple can no longer be an offer containing server-side quality values.
+  - An offer will only match a wildcard clause in the header, such as ``*/*``
+    or ``text/*`` if it does not match something more specific.
```
whiteroses, line 55: If I remember correctly, their behaviour will be modified more than slightly: the switch from the old algorithms to the one in `.acceptable_offers` involved many changes - some of which were documented in several of the docstrings that are being removed, but even those docstrings didn't include everything. As these methods have been deprecated for such a long time, it may be more helpful and accurate/precise to say that the methods have switched to using the algorithm in `.acceptable_offers` to conform to RFC 7231?

(I also had trouble understanding "A tuple can no longer be an offer containing server-side quality values." - is it saying that offers can no longer contain server-side quality values?)

(The same text is used in the docstrings of `.best_match`, `.quality` and `.__contains__`, so this comment also applies to them.)

src/webob/acceptparse.py
```diff
@@ -89,6 +90,17 @@ def _list_1_or_more__compiled_re(element_re):
     )
 
 
+class HeaderState(enum.Enum):
+    """
+    The state of an accept-style header to assist in identifying scenarios
+    an application may want to know about during accept-negotation.
```
whiteroses, line 96: Typo in "accept-negotation"?

The docstring seems a bit unclear - it may be helpful to explain that an Accept header may have a valid value, an invalid value, or may be missing from the request; and that this enum represents which of the three states the header is in?

src/webob/acceptparse.py
```diff
@@ -339,7 +362,7 @@ def _parse_media_type_params(cls, media_type_params_segment):
                 value = cls._process_quoted_string_token(token=value)
                 media_type_params[index] = (name, value)
 
-        return media_type_params
+        return tuple(media_type_params)
```
whiteroses, line 365: The return value has been changed to a tuple, but the docstring is still saying that it returns a list?

src/webob/acceptparse.py
```diff
@@ -358,6 +381,9 @@ def _python_value_to_header_str(cls, value):
         Convert Python value to header string for __add__/__radd__.
```
whiteroses, line 381: Both the class method's name and its docstring are saying that the class method returns a header string, but it has been modified to return `None` when `value` is `None`?

The class method is also used in `.__init__` and `accept_property` (and indirectly in `.__add__` and `.__radd__`, via use of `create_accept_header`), so the docstring saying that it is for `__add__` and `__radd__` is no longer true?

src/webob/acceptparse.py
```diff
@@ -421,7 +447,7 @@ def parse(cls, value):
                  | *extension_params* is the extension parameters, as a list
```
whiteroses, line 447: "extension_params" has been changed from a list to a tuple?

src/webob/acceptparse.py
```diff
@@ -555,10 +592,10 @@ def header_value(self):
     @property
     def parsed(self):
         """
-        (``list`` or ``None``) Parsed form of the header.
+        (``tuple`` or ``None``) Parsed form of the header.
 
-        A list of (*media_range*, *qvalue*, *media_type_params*,
-        *extension_params*) tuples, where
+        A tuple of ``(*media_range*, *qvalue*, *media_type_params*,
+        *extension_params*)`` tuples, where
```
whiteroses, line 598: Same issue as https://github.com/Pylons/webob/pull/460/files#r1531436169 with the asterisks here and in the rest of this docstring.

src/webob/acceptparse.py
```diff
@@ -901,25 +699,35 @@ def acceptable_offers(self, offers):
...
-                       include media type parameters) or pre-parsed instances
-                       of :class:`.AcceptOffer`.
-        :return: A list of tuples of the form (media type, qvalue), in
-                 descending order of qvalue. Where two offers have the same
-                 qvalue, they are returned in the same order as their order in
-                 `offers`.
-        """
-        parsed = self.parsed
+        Any offers that cannot be parsed via :meth:`.parse_offer`
+        will be ignored.
```
whiteroses, line 710: `.parse_offer` is not currently included in the documentation, so this is not linking to anything?

src/webob/acceptparse.py
```diff
@@ -901,25 +699,35 @@ def acceptable_offers(self, offers):
...
+
+        :param offers:
+            ``iterable`` of ``str`` media types (media types can
```
whiteroses, line 713: `offers` is a sequence, not an iterable - I think it is the same issue as #442, and there are probably a few of these around, likely my mistakes from back then. Not sure if you'd like to fix them with these PRs - or I could submit a PR to fix them after you're done.

src/webob/acceptparse.py
```diff
@@ -901,25 +699,35 @@ def acceptable_offers(self, offers):
...
+            include media type parameters) or pre-parsed instances
+            of :class:`.AcceptOffer`.
+        :return:
+            A list of tuples of the form (media type, qvalue), in
+            descending order of qvalue. Where two offers match the same
+            qvalue, they are returned in the same order as their order in
+            ``offers``.
```
whiteroses, line 720: The documentation for the return value from `_AcceptInvalidOrNoHeader.acceptable_offers` needs to be included here to explain what happens with the return value when the header is invalid or missing?

src/webob/acceptparse.py
```diff
@@ -1014,619 +822,97 @@ def acceptable_offers(self, offers):
...
-           This is currently maintained for backward compatibility, and will be
-           deprecated in the future.
+        :param offers:
+            (iterable)
 
-           :meth:`AcceptValidHeader.best_match` uses its own algorithm (one not
-           specified in :rfc:`RFC 7231 <7231>`) to determine what is a best
-           match. The algorithm has many issues, and does not conform to
-           :rfc:`RFC 7231 <7231>`.
+            | Each item in the iterable must be a ``str`` media type and
```
whiteroses, line 834: "iterable" should be "sequence", as explained [here](https://github.com/Pylons/webob/pull/460/files#r1531444937).

src/webob/acceptparse.py
```diff
@@ -1014,619 +822,97 @@ def acceptable_offers(self, offers):
...
+              may contain params/extensions.
```
whiteroses, line 835: Media types do not have extension parameters (https://datatracker.ietf.org/doc/html/rfc7231.html#section-3.1.1.1), and `.acceptable_offers` does not recognise them in media type offers.

src/webob/acceptparse.py
```diff
@@ -1014,619 +822,97 @@ def acceptable_offers(self, offers):
...
-        match. If the two have the same number of '*'s, the one that shows up
-        first in `offers` is the best match.
+        :return:
+            (``str``, or the type of ``default_match``)
 
-        :param offers: (iterable)
+            | The offer that is the best match based on q-value. If there is no
+              match, the value of ``default_match`` is returned. Where two
+              offers match the same qvalue, they are returned in the same order
+              as their order in ``offers``.
```
whiteroses, line 846: `.best_match` returns a single offer, so the "they are returned" seems a little confusing?

Also, inconsistent spelling of "qvalue"/"q-value"? (RFC spells it without a hyphen.)

src/webob/acceptparse.py
```diff
@@ -1014,619 +822,97 @@ def acceptable_offers(self, offers):
...
-        :rfc:`RFC 7231, section 5.3.2 <7231#section-5.3.2>`. It does not
-        correctly take into account media type parameters:
+        :param offer: (``str``) media type offer
+        :return: (``float`` or ``None``)
 
-            >>> instance = AcceptValidHeader('text/html')
-            >>> instance.best_match(offers=['text/html;p=1']) is None
-            True
+                 | The highest quality value from the media range(s) that match
+                   the `offer`, or ``None`` if there is no match.
```
whiteroses, line 873: Would be helpful to copy over the documentation for the return value in `_AcceptInvalidOrNoHeader.quality`, to explain what happens when the header is invalid or missing?

src/webob/acceptparse.py
```diff
@@ -1641,86 +927,54 @@ def __add__(self, other):
...
-        If `other` is a valid header value or an :class:`AcceptValidHeader`
-        instance, then a new :class:`AcceptValidHeader` instance with the valid
-        header value is returned.
-
-        If `other` is ``None``, an :class:`AcceptNoHeader` instance, an invalid
-        header value, or an :class:`AcceptInvalidHeader` instance, a new
-        :class:`AcceptNoHeader` instance is returned.
+        The rules for adding values to a header are that the values are
+        appended if valid, or discarded. If everything is discarded then an
+        instance representing a missing header is returned.
```
whiteroses, line 935: This reads as a bit unclear - maybe could do with some rewording, or the explanations from the docstrings of the three `.__add__`s could be copied over and updated?

src/webob/acceptparse.py
```diff
@@ -1641,86 +927,54 @@ def __add__(self, other):
...
+        is_self_valid = self.header_state is HeaderState.Valid
+        is_other_valid = other.header_state is HeaderState.Valid
 
-        return self._add_instance_and_non_accept_type(instance=self, other=other)
+        if is_self_valid:
+            if is_other_valid:
+                if self.header_value == "":
+                    return other
+                if other.header_value == "":
+                    return self
```
whiteroses, line 947: The docstring states that a new header object is always created, but if `other.header_value == ""`, `self` is returned?

src/webob/acceptparse.py
```diff
@@ -1641,86 +927,54 @@ def __add__(self, other):
...
+        appended if valid, or discarded. If everything is discarded then an
+        instance representing a missing header is returned.
         """
 
-        if isinstance(other, AcceptValidHeader):
-            return AcceptValidHeader(header_value=other.header_value)
-
-        if isinstance(other, (AcceptNoHeader, AcceptInvalidHeader)):
-            return AcceptNoHeader()
+        other = create_accept_header(other)
```
whiteroses, line 938: This is using `create_accept_header` to handle all the possible types for the value of `other`, but `create_accept_header` is documented as taking a parameter of type `str` or `None`?

src/webob/acceptparse.py
```diff
@@ -1641,86 +927,54 @@ def __add__(self, other):
...
-    :param header_value: (``str``) header value
-    :return: If `header_value` is ``None``, an :class:`AcceptNoHeader`
-             instance.
-
-             | If `header_value` is a valid ``Accept`` header, an
-               :class:`AcceptValidHeader` instance.
-
-             | If `header_value` is an invalid ``Accept`` header, an
-               :class:`AcceptInvalidHeader` instance.
+    :param header_value: (``str`` or ``None``) header value
```
whiteroses, line 971: It needs to be mentioned that `None` is used to represent a missing header?

docs/api/webob.txt
```diff
@@ -23,85 +23,71 @@ a certain type:
...
-.. autoclass:: AcceptValidHeader
-   :members: parse, header_value, parsed, __init__, __add__, __bool__,
-	     __contains__, __iter__, __radd__, __repr__, __str__,
-	     accept_html, accepts_html, acceptable_offers, best_match, quality
-
-.. autoclass:: AcceptNoHeader
-   :members: parse, header_value, parsed, __init__, __add__, __bool__,
-	     __contains__, __iter__, __radd__, __repr__, __str__,
-             accept_html, accepts_html, acceptable_offers, best_match, quality
+.. autoenum:: HeaderState
```
whiteroses, line 29: "and their methods" makes the sentence quite ambiguous, and as the documentation for the classes also includes their non-method attributes, perhaps "and their methods" could be removed? (Assuming I'm reading the intended meaning correctly.)

Under this, `AcceptOffer` and `HeaderState` are included - but they are not classes that "may be returned by one of the functions above"?

tests/test_acceptparse.py
```diff
@@ -466,1570 +475,426 @@ def test_parse_offer__invalid(self, offer):
...
 
-    @pytest.mark.parametrize("value", ["", [], (), {}])
-    def test___add___valid_empty_value(self, value):
-        left_operand = AcceptValidHeader(header_value=",\t ,i/j, k/l;q=0.333,")
-        result = left_operand + value
-        assert isinstance(result, AcceptValidHeader)
-        assert result.header_value == left_operand.header_value
-        assert result is not left_operand
+    def test_copy(self):
+        instance = Accept("*/plain;charset=utf8;x-version=1")
```
whiteroses, line 539: `*/plain` is valid according to https://datatracker.ietf.org/doc/html/rfc7231.html#section-5.3.2, but I think this is an oversight in the spec - I don't think they intended for it to be possible for `type` to be `*` when the `subtype` is not `*`. Not 100% certain about this, but as I think we can use any valid header value here, probably better to use one that we know is intended to be valid.

From the RFC: 'The "q" parameter is necessary if any extensions (accept-ext) are present, since it acts as a separator between the two parameter sets.' So this needs a "q" parameter before the extension parameter for the header to be valid (assuming that you want to use a valid header value here, and not an intentionally invalid one).

The missing "q" parameter does not seem like it's being detected by the regex at the moment - however, [extension parameters have been removed in RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-changes-from-rfc-7231), which obsoletes RFC 7231, so whether it's worth fixing may depend on when you would like WebOb to start conforming to RFC 9110?

tests/test_acceptparse.py
```diff
@@ -466,1570 +475,426 @@ def test_parse_offer__invalid(self, offer):
...
-            accept.best_match(["image/Tiff", "image/PnG", "image/jpg"]) == "image/PnG"
-        )
-
-    @pytest.mark.filterwarnings(IGNORE_BEST_MATCH)
-    @pytest.mark.filterwarnings(IGNORE_CONTAINS)
-    def test_best_match_zero_quality(self):
-        assert (
-            AcceptValidHeader("text/plain, */*;q=0").best_match(["text/html"]) is None
-        )
-        assert "audio/basic" not in AcceptValidHeader("*/*;q=0")
```
whiteroses, line 1227: These asserts seem to be worth keeping - I don't think they or any functionally equivalent asserts are in the new tests? (One assert is for `.best_match` and the other is for `.__contains__` - not sure why the `.__contains__` one was placed into a `.best_match` test method.)

tests/test_acceptparse.py
```diff
@@ -2042,71 +907,308 @@ def test_accepts_html(self):
...
+        if request.param == "add":
+            return lambda x, y: x + y
+        return lambda x, y: y + x
+
+    @pytest.mark.parametrize(
+        "input_value, input_header",
+        valid_values_with_headers,
+    )
+    def test_valid_add_missing(self, input_value, input_header, maker, fn):
+        inst = Accept(input_value)
```
whiteroses, line 1062: In the `Accept.__init__` docstring, the `header_value` parameter is documented to be of type `str` or `None`, but other types are passed to `Accept.__init__` in this test and quite a few others?

tests/test_acceptparse.py
```diff
@@ -2042,71 +907,308 @@ def test_accepts_html(self):
...
-        )
+        assert accept.best_match(["audio/basic", "text/html"]) == "audio/basic"
         assert accept.best_match([], default_match="fallback") == "fallback"
 
-    @pytest.mark.filterwarnings(IGNORE_QUALITY)
     def test_quality(self):
-        instance = AcceptInvalidHeader(header_value=", ")
-        returned = instance.quality(offer="type/subtype")
-        assert returned == 1.0
+        accept = Accept(None)
```
whiteroses, line 924: This should be `Accept(", ")` instead of `Accept(None)`?

src/webob/acceptparse.py
```diff
@@ -89,6 +90,17 @@ def _list_1_or_more__compiled_re(element_re):
...
+    """
+
+    Valid = "Valid"
+    Invalid = "Invalid"
+    Missing = "Missing"
+
+
 class AcceptOffer(namedtuple("AcceptOffer", ["type", "subtype", "params"])):
     """
     A pre-parsed offer tuple represeting a value in the format
```
whiteroses, line 106: Typo in "represeting"?

whiteroses (commented):
The "Accept-* headers" section of docs/reference.txt would also need to be updated for this PR ([L347](https://github.com/Pylons/webob/blob/simplify-accept-parsing/docs/reference.txt#L347) and [L358-364](https://github.com/Pylons/webob/blob/simplify-accept-parsing/docs/reference.txt#L358)).

Could use `:type:`  and `:rtype:` (https://www.sphinx-doc.org/en/master/usage/domains/python.html#info-field-lists) to specify parameter and return value types in the documentation?

[RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231.html) is obsoleted by [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110) - does that affect this PR?

Otherwise, all looks good from what I can see!

mmerickel:
@whiteroses Thank you so much - excellent reviews. I will work on these when I'm back from vacation and try to apply similar fixes to the other branches.

tests/test_acceptparse.py
mmerickel: I'm open to removing extension param support - I don't have a strong opinion here. I'm not sure it's really hurting things much to keep them though.

src/webob/acceptparse.py
mmerickel: The add and radd support what they did before - I just moved the logic into the `__init__` as you noted. It sounds like you just want the `__init__` to be documented to take those other types which is fine - I can do that.

whiteroses: If you are actually changing `.__init__` to make it accept all those other types of values accepted by `.__add__`/`.__radd__`, then yes, it's only the documentation that needs to be updated.

tests/test_acceptparse.py
whiteroses: RFC 9110 says that "The accept extension grammar (accept-params, accept-ext) has been removed because it had a complicated definition, was not being used in practice, and is more easily deployed through new header fields." I imagine they must be pretty certain that it's not being used in practice, to feel able to make the change.

I don't see how it would hurt things to keep the extension param support either, but it'd be nice to simplify things and be able to remove some code, I imagine. Not sure what you think about moving to conform to RFC 9110 in general - as RFC 7231 is obsoleted by RFC 9110, I imagine WebOb would want to make the move at some point?

If you would like to keep the extension param support for the time being, I can look into why the regex did not catch the missing "q" parameter separator?

whiteroses:
Yes, most of them were docstring issues - enjoy your vacation!

PR #468, opened 2024-10-10
Fix CI

luhn:
Several Github Actions builds were broken, so I made some changes to get green lights:

MacOS builds were failing because architecture was x64 but `macos-latest` is now M1 machines (arm64).  I decided to remove architecture from the matrix entirely.  x86 is already disabled for Linux and MacOS, it only runs on Windows, so I don't know how useful it is.

Alternative solutions:  Use `macos-13`, which is an x64 runner.  Or expand the architecture matrix so we can run tests across Windows x86, Windows x64, Linux x64, and MacOS arm64.

PyPy3.8 builds were failing because of a segfault during pytests.  This [issue has been resolved](https://pypy.org/posts/2024/03/fixing-bug-incremental-gc.html), but because pypy3.8 is no longer actively supported only pypy3.9 and 3.10 have been fixed.  Because it's no longer supported, I think it's reasonable to remove entirely.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -21,7 +21,6 @@ jobs:
                     - "3.10"
                     - "3.11"
                     - "3.12"
-                    - "pypy-3.8"
                     - "pypy-3.9"
                     - "pypy-3.10"
                 os:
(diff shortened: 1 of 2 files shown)


mmerickel (approved):

miketheman (approved):
Hit this today post-fork update. GHA was sad.

PR #467, opened 2024-10-10
Add Python 3.13 support using `legacy-cgi`

luhn:
A quick fix for #437 while #466 is in progress.

PR #469, opened 2024-10-10
Backport `legacy-cgi` to 1.8 branch

luhn:
Backport of #467

Unfortunate 1.8 is significantly behind the main branch re CI:  Latest tested version is Python 3.7 (!!) and there's no Github Actions.  Doesn't seem much point trying to backport that so I didn't bother.  I did however confirm tests do pass in Python 3.13.

[tox-py313.txt](https://github.com/user-attachments/files/17337305/tox-py313.txt)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,9 @@ setup(
     packages=find_packages('src', exclude=['tests']),
     package_dir={'': 'src'},
     python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*',
+    install_requires=[
+        "legacy-cgi>=2.6; python_version>='3.13'",
+    ],
     zip_safe=True,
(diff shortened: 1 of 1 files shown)


mmerickel (approved):
Yeah the other stuff on main doesn't seem to have to do with python version compatibility so doesn't need to be pulled in here.

I didn't realize github actions aren't enabled on 1.8 though, that's not great. :-( This won't make things any worse tho so it's probably ok to just get it out there.

digitalresistor:
The 1.8-branch is also YEARS old at this point. It only got a recent fix due to it being a security thing. 1.8 technically is still compatible with Python 2...

digitalresistor:
https://pypi.org/project/WebOb/1.8.9/

Issue #473, opened 2024-12-09
Python 3.13 deprecation of utcnow()


jenstroeger:
Hello,

I’m seeing several deprecation warnings like
```
/path/to/.venv/lib/python3.13/site-packages/webob/cookies.py:238: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
  v = datetime.utcnow() + v
```
and other `utcnow()` related calls. Are there any plans to address this issue soon? Happy to provide a PR.

mmerickel:
I have no concerns about any internal fixes - but I'm worried about areas where it'll hit the public api surface.

jenstroeger:
Right… by moving from naive `utcnow()` to aware `now(tz=timezone.utc)` I ended up moving the whole stack to aware datetimes. Probably a long overdue move anyway.

In the case of WebOb I think a transparent change would be e.g. from

https://github.com/Pylons/webob/blob/39d5af3c797e7b867f152c2e8c979de42d029403/src/webob/cookies.py#L241-L242
to
```python
 if isinstance(v, timedelta): 
     v = (datetime.now(tz=timezone.utc) + v).replace(tzinfo=None)
```
This would be a drop-in change to address the deprecation, but could probably be simplified case by case.

Shouldn’t that address your concern about the public API surface?

mmerickel:
I guess I'm roping in the reason for the deprecation - not just how to workaround the warning.

The goal of the deprecation is to get people to stop using naive datetimes and/or moving toward assuming that a naive datetime is "local" tz instead of UTC (because this is consistent with the stdlib apis and how they handle naive datetimes).

So sure, if we are just tackling the warning then yeah you can simply replace usages of `datetime.utcnow()` with `datetime.now(datetime.UTC).replace(tzinfo=None)` and we are done.

I was assuming you were also proposing the rest, in which you're actually changing the library to use tz-aware datetime objects which is a much bigger change.

Can you clarify your proposal?

mmerickel:
To be clear - changing webob to expect tz-aware datetimes and expose tz-aware datetimes at the public api surface is secondary and much more controversial - we'll need to do that carefully.

jenstroeger:
> […] we'll need to do that carefully.

Agreed. But mixing aware and naive datetimes also requires care because they can’t be compared. So I think the question becomes:

[...]

Issue #430, opened 2021-04-23
Setting response.expires via timedelta produces an incorrect header value


MarcSchmitzer:
When using a timedelta to calculate the absolute timestamp, webob uses `datetime.now()`, which is in local time. This is later re-interpreted to be UTC, which yields an "Expires" header value that is off by the UTC offset of the local timezone.

Example in CEST:
```
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime, timedelta
>>> from webob import Response
>>> datetime.utcnow()
datetime.datetime(2021, 4, 23, 13, 48, 1, 541647)
>>> r = Response()
>>> r.expires = timedelta(seconds=0)
>>> r.expires
datetime.datetime(2021, 4, 23, 15, 48, 13, tzinfo=UTC)
```
The example shows that setting `expires` to a zero timedelta produces a header value two hours in the future.

The `cache_expires` method does the right when passed a timedelta.

PR #475, opened 2024-12-12
Replace to-be-deprecated utcnow() datetime calls with aware now() calls

jenstroeger:
As per issue https://github.com/Pylons/webob/issues/473

I didn’t add tests yet, see my comments below. ~Still need to format the code, but it looks like Windows tests fail because, I suspect, of [this data source discrepancy](https://docs.python.org/3.13/library/zoneinfo.html#data-sources).~

### DIFF
--- a/src/webob/cookies.py
+++ b/src/webob/cookies.py
@@ -1,7 +1,7 @@
 import base64
 import binascii
 from collections.abc import MutableMapping
-from datetime import date, datetime, timedelta
+from datetime import date, datetime, timedelta, timezone
 import hashlib
 import hmac
 import json
@@ -239,10 +239,13 @@ def serialize_cookie_date(v):
         v = timedelta(seconds=v)
 
     if isinstance(v, timedelta):
-        v = datetime.utcnow() + v
+        v = datetime.now(tz=timezone.utc) + v
 
-    if isinstance(v, (datetime, date)):
+    if isinstance(v, datetime):
+        v = v.astimezone(timezone.utc).timetuple()
+    elif isinstance(v, date):
         v = v.timetuple()
+
     r = time.strftime("%%s, %d-%%s-%Y %H:%M:%S GMT", v)
 
     return bytes_(r % (weekdays[v[6]], months[v[1]]), "ascii")
--- a/src/webob/response.py
+++ b/src/webob/response.py
@@ -1,5 +1,5 @@
 from base64 import b64encode
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from hashlib import md5
 import re
 import struct
@@ -1246,6 +1246,8 @@ class Response:
             seconds = timedelta_to_seconds(seconds)
         cache_control = self.cache_control
 
+        utcnow = datetime.now(tz=timezone.utc).replace(tzinfo=None)
+
         if seconds is None:
             pass
         elif not seconds:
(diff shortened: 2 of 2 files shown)


src/webob/cookies.py
```diff
@@ -239,10 +240,13 @@ def serialize_cookie_date(v):
         v = timedelta(seconds=v)
 
     if isinstance(v, timedelta):
-        v = datetime.utcnow() + v
+        v = datetime.now(tz=zoneinfo.ZoneInfo("UTC")) + v
```
jenstroeger: The string formatting below doesn’t care that the datetime object is aware here.

src/webob/cookies.py
```diff
@@ -239,10 +240,13 @@ def serialize_cookie_date(v):
...
 
-    if isinstance(v, (datetime, date)):
+    if isinstance(v, datetime):
+        v = v.astimezone(zoneinfo.ZoneInfo("UTC")).timetuple()
+    elif isinstance(v, date):
         v = v.timetuple()
```
jenstroeger: The function can receive `bytes`, `str`, `int` but it doesn’t have an else branch to catch any other unexpected type. If for some reason a datetime is passed in then we’ll want to make sure it’s in UTC, hence the adjustment. Also, unless a date object is passed in there shouldn’t be a need to handle those but cookies should always contain a date & time AFAIK.

src/webob/response.py
jenstroeger, line 1: All `utcnow()` are replaced and continue to produce naive datetime objects in UTC.

jenstroeger (commented):
A few comments below to continue the conversation from issue https://github.com/Pylons/webob/issues/473.

mmerickel:
`datetime.timezone.utc` was added in python 3.2 - is there a reason not to use it in favor of zoneinfo?

jenstroeger:
> `datetime.timezone.utc` was added in python 3.2 - is there a reason not to use it in favor of zoneinfo?

Sigh 🤦🏻‍♂️ Consider that a magnificent brain fart on my side.

mmerickel (approved):
LGTM

jenstroeger:
@mmerickel what do you think of the above comments/questions, esp https://github.com/Pylons/webob/pull/475#discussion_r1881368747?

src/webob/cookies.py
mmerickel: This change would require a test. For the specific utcnow issue I think #480 is probably a good way to go. But this is a good point that isn't handled anywhere else.

digitalresistor:
Thank you for this, and for the patience on the long review cycle here.

There were three overlapping PRs open against #473 and #430 (this one, #480 and #491), all touching the same lines, so none could be merged alongside the others. I've combined them into #496, where your commit is preserved with you as the author.

Your change to convert an aware `datetime` to UTC before serializing it carries over as its own bugfix — as @mmerickel noted in review, that's a real problem not handled anywhere else. It now has the test he asked for.

One adjustment worth flagging: the `astimezone(timezone.utc)` call is gated on `v.tzinfo is not None` in #496. Applied unconditionally it treats a *naive* datetime as local time and shifts it by the machine's UTC offset, where WebOb has always read naive datetimes as UTC. Under `TZ=America/New_York` a naive `datetime(2011, 1, 4, 13, 43, 50)` serialized as `18:43:50 GMT` instead of `13:43:50 GMT`. CI runs in UTC, so it doesn't show up there. There's a tz-sensitive regression test for it now.

For the deprecation itself #496 uses the centralized `utcnow()` helper from #480, which @mmerickel favoured in his review of this PR.

PR #476, opened 2024-12-13
UTF-8 encoded data should not by encoded with latin-1

jmadotgg:
I had problems non ascii letters like umlauts and ß in the requests url path, even though they where uri encoded and also decoded properly before the latin-1 encode step in line 120.
I know this might be a naive fix, maybe you can come up with a better overall approach or guide me towards it. I also don't know if utf-8 encoded uris are out of scope for this library.

### DIFF
--- a/src/webob/request.py
+++ b/src/webob/request.py
@@ -114,7 +114,7 @@ class BaseRequest:
             return val
         encoding = getattr(self, encattr)
 
-        if encoding in _LATIN_ENCODINGS:  # shortcut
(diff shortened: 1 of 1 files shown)


mmerickel:
Can you show an example url and a stack trace?

generally speaking a utf8 path that is properly encoded will work fine with webob. 

The Latin-1 encoding is used within PEP3333 to marshal bytes around as Unicode code points but they are rarely interpreted that way. That being said in HTTP a lot of things like headers don’t support utf8.

jmadotgg:
@mmerickel Thank you for your quick response, I will get back to this the upcoming week.

Issue #457, opened 2023-12-06
update or remove `__all__` declarations


jvanasco:
Many of the files in WebOb declare `__all__` for exports.  This can create problems with the current generation of type checkers, as they may not be able to see functions that do not appear in these declarations.  For example, in `webob.cookies`, `parse_cookie` is not declared.

The `__all__` declarations are only used for * imports - which are increasingly unused in exchange for explicit imports.  

I would be happy to generate a PR for either situation, however I would prefer dropping `__all__`.

digitalresistor:
Thanks for raising this. I went through the history and current state of each module before answering. Short version: I'd rather not drop `__all__`, the `parse_cookie` example is working as intended, but you've landed near a real gap — just not the one you cited. Details, since I think the underlying goal is worth getting right.

## `__all__` isn't the lever today — but #477 makes it one

WebOb currently ships no `py.typed` marker and no inline annotations, so nothing in `src/webob/*.py` reaches mypy or pyright. Type checkers resolve `webob` through typeshed's `types-WebOb` stubs (1.8.0.20260716), which carry their own `__all__`. Editing these lists today has no effect on them.

That changes with #477 (tracked by #474), which adds `src/webob/py.typed` and wires it into `setup.py` via `package_data`. Once that lands, the re-export rules become live for real consumers.

Worth separating the two mechanisms, though: `__all__` governs re-export of names a module **imports**, not visibility of names a module **defines**. `parse_cookie` is defined at `src/webob/cookies.py:208`, so it resolves fine either way. Your specific example wouldn't be fixed by dropping or extending `__all__`.

## Why `parse_cookie` isn't exported

It was never intended as public API. It landed in d484fa2 (Sept 2011) as part of *"change req.cookies implementation to parse on demand"* — a helper extracted to serve that refactor. `cookies.__all__` started as `['Cookie']` in 6a83d65 (2010) and has been curated deliberately since. In the fifteen years after, `parse_cookie` has picked up exactly one call site in the library (`src/webob/cookies.py:51`) and no documentation.

I'd also rather not freeze its current signature. It accepts a `str`, latin-1 encodes it (`_parse_cookie`, `src/webob/cookies.py:201-205`), and yields `(bytes, bytes)` pairs — which its sole caller immediately utf-8 decodes both halves of. That asymmetry is an implementation detail. If there's demand for a public cookie-header parser, I'd rather add a properly-shaped `str` → `str` function than promote this one; happy to take that as its own issue.

[...]

PR #477, opened 2025-02-10
mypy: Add type hints

### LINKED ISSUE #474: Adding type hints
As mentioned in Pyramid's issue [#2638](https://github.com/Pylons/pyramid/issues/2638), I would like to start adding type annotations to WebOb.

Since I don't really know the code base really well _(I'm learning on the go!)_, I'm planning to make many little PRs so that they are easily reviewed and get feedbacks quickly (also because they should not affect runtime at all).

Can I get a green light on this?

Also, is it okay to make use of this issue to track progress? Google group is too noisy for my liking.

Daverball:
Closes: #474

My curiosity got the better of me. So this is the stubs I contributed to typeshed automatically merged into the source using `merge_pyi`.

Afterwards I cleaned up any mistakes and completed the type hints for the internal API. I added a couple of type tests, but we could definitely add more.

I tried to keep changes to the actual code to a bare minimum. I did however change a couple of minor things. Like remove an unnecessary check for Python 2.7

You can do with these what you want, the diff is fairly large, so it will take some time to review.


jvanasco:
I think found a typo in your stubs-

You define a Protocol here:  https://github.com/python/typeshed/blob/main/stubs/WebOb/webob/cookies.pyi#L29-L31

```python
class _Serializer(Protocol):
    def loads(self, appstruct: Any, /) -> bytes: ...
    def dumps(self, bstruct: bytes, /) -> Any: ...
```

I think that should read:
```python
class _Serializer(Protocol):
    def dumps(self, appstruct: Any, /) -> bytes: ...
    def loads(self, bstruct: bytes, /) -> Any: ...
```

It's basically just swapping the function names.

Notice how the input/output are reversed of the other functions: https://github.com/python/typeshed/blob/main/stubs/WebOb/webob/cookies.pyi#L96-L104

```python
class JSONSerializer:
    def dumps(self, appstruct: Any) -> bytes: ...
    def loads(self, bstruct: bytes | str) -> Any: ...

class Base64Serializer:
    serializer: _Serializer
    def __init__(self, serializer: _Serializer | None = None) -> None: ...
    def dumps(self, appstruct: Any) -> bytes: ...
    def loads(self, bstruct: bytes) -> Any: ...
```

Note that `SignedSerializer` uses `JSONSerializer` if no `serializer` is supplied:

https://github.com/Pylons/webob/blob/main/src/webob/cookies.py#L705-L706

And `CookieProfile` uses the `Base64Serializer`

https://github.com/Pylons/webob/blob/main/src/webob/cookies.py#L813-L814

Daverball:
@jvanasco Good catch. Feel free to open an issue or PR on typeshed.

src/webob/cookies.py
```diff
@@ -1,17 +1,55 @@
...
+    # we accept both the official spelling and the one used in the WebOb docs
+    # the implementation compares after lower() so technically there are more
+    # valid spellings, but it seems more natural to support these two spellings
+    _SameSitePolicy: TypeAlias = Literal[
+        "Strict", "Lax", "None", "strict", "lax", "none"
+    ]
+
+    class _Serializer(Protocol):
+        def loads(self, appstruct: Any, /) -> bytes: ...
+        def dumps(self, bstruct: bytes, /) -> Any: ...
```
Daverball:
```suggestion
        def dumps(self, appstruct: Any, /) -> bytes: ...
        def loads(self, bstruct: bytes, /) -> Any: ...
```

digitalresistor:
Following up from #457, which I closed in favour of this PR — thanks for putting this together, having the typeshed stubs merged back inline is a big step.

One gap worth flagging: this PR adds `src/webob/py.typed` but doesn't touch `src/webob/__init__.py`, which still imports six names for re-export that aren't in its `__all__`:

```
BaseRequest, parse_date, parse_date_delta,
serialize_date, serialize_date_delta, timedelta_to_seconds
```

Harmless today, since nothing reads our inline types. With `py.typed` shipped it becomes a downstream error. Installing this branch into a clean venv and checking a consumer that does `from webob import BaseRequest, parse_date`:

```
# pyright, default settings — reportPrivateImportUsage is an error at basic level
error: "BaseRequest" is not exported from module "webob"
  Import from "webob.request" instead  [reportPrivateImportUsage]

# mypy --strict
error: Module "webob" does not explicitly export attribute "BaseRequest"  [attr-defined]
```

All six fail the same way. Plain `mypy` without `--strict` is clean, so the blast radius is every pyright user plus strict mypy users.

The part that makes it easy to miss: **our own CI won't catch it.** `mypy -p webob` passes clean on this branch (19 files, no issues), because `no_implicit_reexport` flags the *importing* module and nothing inside WebOb imports these from the top-level package. It only surfaces for consumers, after release.

Nothing for you to do about the `__init__.py` side — I've handled it in #497, since it's an API decision rather than a typing one. Since 2.0 is a major, we're dropping the five date helpers from the top-level namespace outright rather than exporting them: they moved to `webob.datetime_utils` in 0.9.7 (0196b09) and have only been re-exported from `webob` for backwards compatibility since, never documented and never in `__all__`. `BaseRequest` is genuinely public, so it just gets added to `__all__`. The `UTC` tzinfo and the `year`/`month`/`week`/`day`/`hour`/`minute`/`second` constants are unaffected.

With #497 applied on top of this branch, the same consumer file is clean under both pyright and `mypy --strict`.

One thing I would like to ask of you here, though: the `tests/mypy/` harness you added catches this class of bug nicely, and a permanent check there would stop top-level re-exports silently regressing again. A file containing just the public top-level imports is enough — I tried it against `mypy tests/mypy/` and it failed as expected before the fix:

```
tests/mypy/check_reexports.py:3: error: Module "webob" does not explicitly export attribute "BaseRequest"
tests/mypy/check_reexports.py:3: error: Module "webob" does not explicitly export attribute "parse_date"
```

It makes more sense on this branch than on #497, which has no mypy dependency to hang it off — so I've left it out of that PR. Note it'll need `BaseRequest` importable from `webob` and the date helpers from `webob.datetime_utils` to match #497's end state.

Daverball:
I think I've made all of the changes you asked for and resolved the merge conflicts.

I went ahead and already added `BaseRequest` to `__all__` in `__init__.py` so the new test passes, but I left out the other changes from #497.

Daverball:
I guess the web-based merge was too ambitious, I will have to fix this on my other computer, when I find the time.

PR #480, opened 2025-06-09
Replace deprecated datetime.datetime.utcnow

kajinamit:
The utcnow function was deprecated in Python 3.12[1].

Note that all datetime instances are kept non-timezone-aware to keep backword-compatibility.

[1] https://docs.python.org/3.13/library/datetime.html#datetime.datetime.utcnow

### DIFF
--- a/src/webob/datetime_utils.py
+++ b/src/webob/datetime_utils.py
@@ -1,5 +1,5 @@
 import calendar
-from datetime import date, datetime, timedelta, tzinfo
+from datetime import date, datetime, timedelta, timezone, tzinfo
 from email.utils import formatdate, mktime_tz, parsedate_tz
 import time
 
@@ -19,6 +19,7 @@ __all__ = [
     "serialize_date",
     "parse_date_delta",
     "serialize_date_delta",
+    "utcnow",
 ]
 
 _now = datetime.now  # hook point for unit tests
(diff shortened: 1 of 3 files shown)


piotr-dobrogost:
Related issue #473

miketheman:
Duplicate of #475 and does a lot more - looks like you ran an auto-formatter or something changed a lot more lines than is relevant?

kajinamit:
> Duplicate of #475 and does a lot more - looks like you ran an auto-formatter or something changed a lot more lines than is relevant?

That was due to the base commit to fix the lint check error. I removed that commit, because it was merged separately.

src/webob/datetime_utils.py
mmerickel, line 21: I think utcnow should be added to `__all__` since this module is following that pattern.

kajinamit: Added it to the list.

digitalresistor:
Thank you for this, and apologies for how long it sat.

There were three overlapping PRs open against #473 and #430 (this one, #475 and #491), all rewriting the same lines, so none could be merged alongside the others. I've combined them into #496, where your commit is preserved with you as the author.

Your approach is the base of that branch: the `webob.datetime_utils.utcnow()` helper returning a naive UTC datetime, which is what @mmerickel favoured in his review of #475 precisely because it keeps the public API surface unchanged. Nothing WebOb exposes changes type, which leaves the tz-aware migration discussed in #473 as a separate piece of work.

Your commit is unmodified. It also ended up resolving #430: `_now` now points at your helper, which fixes the naive-local-time bug in both `serialize_date()` and `parse_date_delta()` at once.

PR #485, opened 2026-01-14
Fix test to stop lint error.

ianjosephwilson:
Seems that the linter was catching this by trying to strip the parens and its blocking the docs build.
Looks like the test itself was incorrectly a no-op in the first place?


digitalresistor:
I merged https://github.com/Pylons/webob/pull/478 which fixed this too, thank you!

